PR #18, opened 2020-05-21
Update HOMEASSISTANT.md

haberda:
My add-on seems to work well after some testing. If you want to include it in the install instructions please feel free. Thanks again.


bbernhard:
Thanks! :)

PR #20, opened 2020-06-12
Add Endpoint for link to existing device

### LINKED ISSUE #9: Feature link devices
It would be great to support the feature of linking devices so we can use my number.

The signal-cli already supports this. So I enter on the docker container instance.
`docker exec -it -u 0 NameOfContainer bash`
and generate the link code `signal-cli link -n "signal-laptop"`
It generate +- a String like this `tsdevice:/\?uuid=asdfghjkl\&pub_key=qwertyuiop`

Then I generate a picture (QR-Code) with https://linux.die.net/man/1/qrencode
`qrencode -o linkqr.png tsdevice:/\?uuid=asdfghjkl\&pub_key=qwertyuiop`
Which I scan with my Android device

After that, I was able to send messages via `signal-cli` from inside of the container instance. (Too good to be true.) When I call the API (`signal-cli-rest-api`) it did not work as I wanted, it gave some errors.

I never tried to write in go. So I don't know if I'm going to try to do this.
Also, I don't really need it. Was just playing around here.

This API makes really easy to integrate other stuff.
Thanks @bbernhard

danijelst:
Add the endpoint /v1/link/:device_name to be able to link to an existing device by generating a tsdevice URI.

With this URI a QR code can be generated (not part of end point) and can be scanned with the existing device to connect them.

Resolves #9

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -257,23 +257,6 @@ func main() {
 		c.JSON(200, about)
 	})
 
-	router.POST("/v1/link/:device_name", func(c *gin.Context) {
-		deviceName := c.Param("device_name")
-		if deviceName == "" {
-			c.JSON(400, gin.H{"error": "Please provide a name for the device"})
-			return
-		}
-
-		command := []string{"--config", *signalCliConfig, "link", "-n", deviceName}
-
-		out, err := runSignalCli(false, command)
-		if err != nil {
-			c.JSON(400, gin.H{"error": err.Error()})
-			return
-		}
-		c.JSON(200, gin.H{"uri": string(out)})
-	})
-
 	router.POST("/v1/register/:number", func(c *gin.Context) {
 		number := c.Param("number")
 
@@ -498,10 +481,23 @@ func main() {
 		}
 	})
 
-	router.GET("/v1/qrcode/:tsdevice_link", func(c *gin.Context) {
-		deviceLink := c.Param("tsdevice_link")
+	router.GET("/v1/qrcodelink", func(c *gin.Context) {
+		deviceName := c.Query("device_name")
 
-		q, err := qrcode.New(deviceLink, qrcode.Medium)
+		if deviceName == "" {
+			c.JSON(400, gin.H{"error": "Please provide a name for the device"})
+			return
+		}
+
+		command := []string{"--config", *signalCliConfig, "link", "-n", deviceName}
+
+		tsdeviceLink, err := runSignalCli(false, command)
+		if err != nil {
+			c.JSON(400, gin.H{"error": err.Error()})
+			return
+		}
+
+		q, err := qrcode.New(string(tsdeviceLink), qrcode.Medium)
 		if err != nil {
 			c.JSON(400, gin.H{"error": err.Error()})
 		}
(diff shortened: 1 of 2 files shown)


src/main.go
```diff
@@ -496,5 +498,23 @@ func main() {
...
+			c.JSON(400, gin.H{"error": err.Error()})
+		}
+
+		q.DisableBorder = true
+		var png []byte
+		png, err = q.PNG(256)
+		if err != nil {
+			c.JSON(400, gin.H{"error": err.Error()})
+		}
+
```
bbernhard, line 515: Many thanks for adding the `qrcode` endpoint - that's awesome!

But I think there's still an issue with url encoding. As the tsdevice link (e.g: `/tsdevice:?uuid=ZiVIV..`) contains special characters that need to url encoded, it currently doesn't work to just copy/paste the `uri` output directly into the browser to get the QR code. (If you would do so so, the `deviceLink` variable (in line 502) just contains the string `tsdevice` without the subsequent `:?uuid=.ZIVIV..`

I guess one way to solve this would be to either directly return the `uri` as base64 encoded string or alternatively add another member to the json object which contains the base 64 encoded uri. That way, the user could directly copy/paste the base64 encoded uri into the qrcode endpoint without worrying about URL encoding. In the qrcode endpoint the base64 encoded tsdevice string would then be decoded before the qrcode gets generated.

(I am doing something similar in the groups endpoint. When listing all the Signal Messenger groups, the groups endpoint returns a `id` and a `internal_id` for every group. The `internal_id` is generated by `signal-cli`/the signal server and contains special characters that would need proper URL encoding when used directly in an URL. Therefore, I introduced the `id`, which (more or less) is just a base64 encoded `internal_id`.)

danijelst: What do you think about a different approach by not having a completely separated endpoint but use the same endpoint which generates the json and append it with a method.

/v1/link/:device_name/qrcode

So those who want just to have the URI would call the /v1/link/:device_name and those which would like to have the QR code, directly the new one. 

This would mean not having to care about the encoding at all and give the end user the possibility to get the QR code with only one call instead of two calls.

bbernhard: Yeah, also thought about this - that would definitely be my preferred solution. My only argument against it would be, that a GET request for creating something could maybe be a bit unexpected? 

I am not sure how the linking of devices is implemented by Signal Messenger, but in my mental model a new resource is created, which would translate more naturally to a POST request IMHO. 

Another thing I've noticed: Signal seems to be creating a different tsdevice uri every time the function is called. e.g: if you use the following request twice, a different uri would be created every time: `curl -X POST -H "Content-Type: application/json" 'http://127.0.0.1:8080/v1/link/test`. 

Not sure if this has any operational implications, but I guess users should probably be aware that they get a different qrcode every time they do a `GET /v1/link/:device_name/qrcode`.

Another option would be to use something like: `GET /v1/qrcodelink?name=<name>`...not sure if this makes it better. 

But I do not hold a strong opinion on either solution...I guess it also depends how one sees the linking.

bbernhard:
Many thanks for your contribution! I'll push a new version to dockerhub in the next days :)

Zixim:
@bbernhard "Add the endpoint /v1/link/:device_name to be able to link to an existing device"

How can we find the device_name to be used in the endpoint ?

PR #36, opened 2020-10-30
Support custom "timeout" option when receiving messages

stayradiated:
Thanks for creating this awesome project, it's made interacting with `signal-cli` so easy!

For some reason, when I query the receive endpoint, no messages are received. 
However If I pass a longer timeout of 5 seconds than I can receive messages.

I've added a `"timeout"` query param to the receive endpoint to allow for a custom timeout value.

For example:

```shell
http -v GET 'http://localhost:8080/v1/receive/+4912345678?timeout=5'
```


bbernhard:
Many thanks for the kind words & thanks a lot for the PR! :+1:

PR #40, opened 2020-11-22
Fix issue when message looks like a CLI arg (like "--test")

fl42:
Hello,

If the message looks like a CLI argument ("-test", "--test", etc) the API call fails.

## How to reproduce
```
$ curl -d '{"message": "--test", "number": "xxx", "recipients": ["xxx"]}' http://127.0.0.1:8080/v2/send
{"error":"usage: signal-cli send [-h] [-g GROUP] [-m MESSAGE]\n                  [-a [ATTACHMENT [ATTACHMENT ...]]] [-e]\n                  [recipient [recipient ...]]\nsignal-cli: error: argument -m/--message: expected one argument\n"}
```

## Proposed fix
Pass the message using stdin ([officially supported by signal-cli](https://github.com/AsamK/signal-cli/blob/1e0aa8929d059b0740da496f6f4ac62159e75141/src/main/java/org/asamk/signal/commands/SendCommand.java#L31))

### DIFF
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -90,7 +90,7 @@ func cleanupTmpFiles(paths []string) {
 
 func send(c *gin.Context, attachmentTmpDir string, signalCliConfig string, number string, message string,
 	recipients []string, base64Attachments []string, isGroup bool) {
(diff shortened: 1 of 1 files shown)


bbernhard:
good catch! Many thanks for the PR :)

Issue #31, opened 2020-10-04
Use non-root user in Dockerfile


ryanjdillon:
It would be nice if the Dockerfile was configured to use a [non-root user](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user) to improve security.

bbernhard:
I've reverted the change for now, as it created backwards compatibility issues with existing bind mounts: https://github.com/bbernhard/signal-cli-rest-api/commit/136fa6269815b1155a1fa5804dcfd4701f15e651

bbernhard:
Thanks to @fl42 there's now a proposal to run `signal-cli-rest-api` without a root user. The way it is implemented, it should work transparently without breaking existing setups. I've already tested the image on my x86-64 and armv7 machines without any issues. 

But it would be great if some more people could give the image a try. The docker image has the tag `bbernhard/signal-cli-rest-api:0.31-dev`. The changes can be found here: https://github.com/bbernhard/signal-cli-rest-api/releases/tag/0.29

I am notifying everybody that has contributed something to this project. If you have time and want to give the image a quick try, it would be highly appreciated. If no issues are raised, the image will become public in the next days :) 


@danijelst @funkybunch @stayradiated @haberda

haberda:
After a little testing this seems to work fine in terms of the HassIO add-on. Thanks!

bbernhard:
> After a little testing this seems to work fine in terms of the HassIO add-on. Thanks!

Many thanks for testing! 

As Christmas and New Year's Eve is probably a busy time for most people, I'll wait another 2-3 days in case someone wants to give it a try. If no issues are raised by then I'll push the new image to dockerhub.

bbernhard:
The image on dockerhub is now updated :tada:. Thanks to @fl42 for implementing the feature and @haberda for testing!

PR #41, opened 2020-11-22
Run as non-root user in Docker

fl42:
Hello,

As mentioned in #31, running as non-root is a best practice.

However current users have a mount point for config with files owned by root.
This fix takes care of updating file ownership on startup.

However with this fix the binary signal-cli-rest-api runs as a child of su process.
Even if "it works", I don't know if there are hidden side effects (signal forwarding?)

Proposed roadmap:
1. Deploy this fix
2. Wait for users to move to this version (thus the owner of their config files will be updated when upgrading to this version)
3. Move back to `ENTRYPOINT ["signal-cli-rest-api"]` with `USER signal-api` directive and remove the entrypoint.sh

What is your opinion on this fix proposal?

I open the PR but please do not merge until we are sure that everything works as expected for everyone (especially current users)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,8 +41,12 @@ RUN cd /tmp/signal-cli-rest-api-src && swag init && go build
 # Start a fresh container for release container
 FROM adoptopenjdk:11-jre-hotspot
 
+RUN groupadd -g 1000 signal-api \
+	&& useradd -M -d /home -s /bin/bash -u 1000 -g 1000 signal-api
+
 COPY --from=buildcontainer /tmp/signal-cli-rest-api-src/signal-cli-rest-api /usr/bin/signal-cli-rest-api
 COPY --from=buildcontainer /tmp/signal-cli /opt/signal-cli
+COPY entrypoint.sh /entrypoint.sh
 
 RUN ln -s /opt/signal-cli/bin/signal-cli /usr/bin/signal-cli
 RUN mkdir -p /signal-cli-config/
@@ -50,4 +54,4 @@ RUN mkdir -p /home/.local/share/signal-cli
 
 EXPOSE 8080
 
-ENTRYPOINT ["signal-cli-rest-api"]
+ENTRYPOINT ["/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,16 +41,15 @@ RUN cd /tmp/signal-cli-rest-api-src && swag init && go build
 # Start a fresh container for release container
 FROM adoptopenjdk:11-jre-hotspot
 
-RUN groupadd -g 1000 signal-api \
-	&& useradd -M -d /home -s /bin/bash -u 1000 -g 1000 signal-api
-
 COPY --from=buildcontainer /tmp/signal-cli-rest-api-src/signal-cli-rest-api /usr/bin/signal-cli-rest-api
 COPY --from=buildcontainer /tmp/signal-cli /opt/signal-cli
 COPY entrypoint.sh /entrypoint.sh
 
-RUN ln -s /opt/signal-cli/bin/signal-cli /usr/bin/signal-cli
-RUN mkdir -p /signal-cli-config/
-RUN mkdir -p /home/.local/share/signal-cli
+RUN groupadd -g 1000 signal-api \
+	&& useradd -M -d /home -s /bin/bash -u 1000 -g 1000 signal-api \
+	&& ln -s /opt/signal-cli/bin/signal-cli /usr/bin/signal-cli \
+	&& mkdir -p /signal-cli-config/ \
+	&& mkdir -p /home/.local/share/signal-cli
 
 EXPOSE 8080
 
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -6,5 +6,11 @@ set -e
 # Fix permissions to ensure backward compatibility
 chown 1000:1000 -R /home/.local/share/signal-cli
 
-# Start API
+# Show warning on docker exec
+cat <<EOF >> /root/.bashrc
+echo "WARNING: signal-cli-rest-api runs as signal-api (not as root!)" 
+echo "Run 'su signal-api' before using signal-cli!"
+EOF
+
+# Start API as signal-api user
 exec su -s /bin/sh -c "exec signal-cli-rest-api" signal-api
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+set -x
+set -e
+
+# Fix permissions to ensure backward compatibility
+chown 1000:1000 -R /home/.local/share/signal-cli
+
+# Start API
+exec su -s /bin/sh -c "exec signal-cli-rest-api" signal-api
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -13,4 +13,4 @@ echo "Run 'su signal-api' before using signal-cli!"
 EOF
 
 # Start API as signal-api user
-exec su -s /bin/sh -c "exec signal-cli-rest-api $@" signal-api
+exec setpriv --reuid=1000 --regid=1000 --init-groups --inh-caps=-all signal-cli-rest-api $@
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -13,4 +13,4 @@ echo "Run 'su signal-api' before using signal-cli!"
 EOF
 
 # Start API as signal-api user
-exec su -s /bin/sh -c "exec signal-cli-rest-api" signal-api
+exec su -s /bin/sh -c "exec signal-cli-rest-api $@" signal-api


bbernhard:
Thanks a lot for your contributions - very much appreciated!

I would see two smaller issues with this approach: 

* Unfortunately, the entrypoint is used in [this](https://github.com/haberda/hassio_addons/blob/master/signal/Dockerfile#L3) Home Assistant addon. I am not sure how many Home Assistant addon users are actually out there, but I already got a few bug reports from addon users - so I think people are using the addon. But I think that would be relatively easy to fix, by changing the go binary's name to something else and then renaming your `entrypoint.sh`script to  `signal-cli-rest-api`. The new entrypoint script would then just need to pass the parameters on to the go binary. I think that could work. 

* How can we make sure that all (or at least the majority of users) have moved to the new version, to revert the temporary fix? I think the majority of signal-cli-rest-api users are using the docker container in conjunction with Home Assistant. And I guess a bunch of those people are probably less tech-savvy people that set up the docker container once and then run it for months (maybe even years without updating). So, if we deploy that fix, let it in there for a few months and then remove it again, I guess it could break those really old setups when those guys finally pull the latest version.

Apart from your suggestion, I also see those options: 

* Push one final  docker image with the tag ``bbernhard/signal-cli-rest-api:latest` to dockerhub, where we log a warning every time a REST call is made. Maybe something like `WARNING: This image is deprecated and won't receive another update. Please read this instructions here on how to perform the upgrade <insert link>`. Then we rename the image and push it to dockerhub with a different name.

That of course won't save the issue when someone pulls in the latest changes via git and manually builds the docker image. But I guess those users are more technically advanced and can fix the permission issues themselves (we could of course also add some additional info to the README.md file to warn existing users). 

* Create a new Dockerfile and maintain two versions (one that runs as root and one that runs as restricted user). Personally I am not a big fan of that, as it adds additional maintenance & testing effort. And as the root version is bad from a security perspective, I think we should get rid of it if possible.

* Break backwards compatibility. That for sure requires the least amount of implementation effort. But I always hate it, if a package upgrade breaks a working setup. So, I am not a big fan of that option and would avoid it at all costs. 

What do you think?

fl42:
Thank you for you complete answer!

> Unfortunately, the entrypoint is used in this Home Assistant addon. I am not sure how many Home Assistant addon users are actually out there, but I already got a few bug reports from addon users - so I think people are using the addon. But I think that would be relatively easy to fix, by changing the go binary's name to something else and then renaming your entrypoint.shscript to signal-cli-rest-api. The new entrypoint script would then just need to pass the parameters on to the go binary. I think that could work.

Another option is to update the Dockerfile of the addon to reflect the upstream change?

> How can we make sure that all (or at least the majority of users) have moved to the new version, to revert the temporary fix? I think the majority of signal-cli-rest-api users are using the docker container in conjunction with Home Assistant. And I guess a bunch of those people are probably less tech-savvy people that set up the docker container once and then run it for months (maybe even years without updating). So, if we deploy that fix, let it in there for a few months and then remove it again, I guess it could break those really old setups when those guys finally pull the latest version.

AFAIK it's hard to have feedback of how many users are using a specific version only using Docker Hub (like the number of downloads by tag)

By temporary fix I mean... few months or even years. If the software runs without issue with the "fix", maybe the fix can become almost permanent.
Or a smart contributor could help us to get ride of using su to change user while maintaining backward compatibility.

> Push one final docker image with the tag ``bbernhard/signal-cli-rest-api:latestto dockerhub, where we log a warning every time a REST call is made. Maybe something likeWARNING: This image is deprecated and won't receive another update. Please read this instructions here on how to perform the upgrade `. Then we rename the image and push it to dockerhub with a different name.

Tech-friendly users will see the warning and act accordingly, but less technical users won't probably see the warning and will stay with a depreciated version forever.

Anyway the proposed fix should not cause any problem for existing users (this has to be confirmed by testing before deploying) even if they upgrade without tech knowledge. The issue is only if they upgrade after we removed the logic to fix permissions.

> That of course won't save the issue when someone pulls in the latest changes via git and manually builds the docker image. But I guess those users are more technically advanced and can fix the permission issues themselves (we could of course also add some additional info to the README.md file to warn existing users).

Totally agree

> Create a new Dockerfile and maintain two versions (one that runs as root and one that runs as restricted user). Personally I am not a big fan of that, as it adds additional maintenance & testing effort. And as the root version is bad from a security perspective, I think we should get rid of it if possible.

As you mentioned this solution adds extra cost that could probably be avoided

> Break backwards compatibility. That for sure requires the least amount of implementation effort. But I always hate it, if a package upgrade breaks a working setup. So, I am not a big fan of that option and would avoid it at all costs.

Totally agree too
Especially for this kind of software where people rely on it for notification.

In conclusion, the proposed fix does not break backward compatibility. We should let the fix in place as long as we want users to be able to migrate automatically.
If using su to start the API as user is not a (too) bad pattern, we can use this solution and leave the logic in place to migrate former users as long as necessary.

Please wait few days in case other people have feedback on this proposal
I think about another solution using setuid, but not as simple as this one.

bbernhard:
> By temporary fix I mean... few months or even years. If the software runs without issue with the "fix", maybe the fix can become almost permanent.

To be honest, I don't really know how Home Assistant addons work. I always thought that when you add a third party addon to Home Assistant, Home Assistant will download the addon definition (i.e Dockerfile) to the system. So, in case you have the ["old Dockerfile"](https://github.com/haberda/hassio_addons/blob/master/signal/Dockerfile) downloaded on your Home Assistant instance, which still uses the old entrypoint and you rebuild the image with the latest `signal-cli-rest-api` image it won't use the new entrypoint. And the docker container can't start due to permission problems. 

> By temporary fix I mean... few months or even years. If the software runs without issue with the "fix", maybe the fix can become almost permanent.

I think it will become permanently - at least that's my experience with those type of fixes ;-). But I don't think that's necessarily bad. I've researched a bit and haven't found anything that speaks against `su -c` inside a docker container. So if we can't come up with something better, I think we could give that a try. But I agree, we maybe should wait a few days to give others a chance to give some feedback too.

fl42:
> To be honest, I don't really know how Home Assistant addons work. I always thought that when you add a third party addon to Home Assistant, Home Assistant will download the addon definition (i.e Dockerfile) to the system. So, in case you have the "old Dockerfile" downloaded on your Home Assistant instance, which still uses the old entrypoint and you rebuild the image with the latest signal-cli-rest-api image it won't use the new entrypoint. And the docker container can't start due to permission problems.

I don't know a lot about hass addons but if they work as you described that should continue to work.
The ENTRYPOINT will be overwritten and the API will run as root, as before?

fl42:
PR opened to update addon: haberda/hassio_addons#9

fl42:
@bbernhard Ok to merge?

bbernhard:
Sorry, wasn't sure whether the PR is already done :) 

Haven't had time to look at it, but I'll have a look in the next days. Do you think it makes sense to push the image first with the `dev` tag (`bernhard/signal-cli-rest-api:latest-dev`) and ask whether a few users (contributors) want to try it out?

fl42:
Oups... I forgot the DO NOT MERGE prefix but actually I think the PR is complete.

> I'll have a look at it in the next days. Do you think it makes sense to push the image first with the dev tag (bernhard/signal-cli-rest-api:latest-dev) and ask whether a few users (contributors) want to try it out?

Sure, I think this is the way to go

bbernhard:
Thanks, looks good to me! :tada: 

The only thing I've noticed is that the `setpriv` is missing in the `adoptopenjdk:11-jdk-hotspot-bionic` docker image, but that's easy to fix :) 

As you might see in the repository history, I've merged your PR and later reverted it again. That was an accident and not intended! I had to revert to signal-cli version 0.6.12 temporarily (as version 0.7.0 introduced some breaking changes which needs some adjustments in the Dockerfile) and accidentally reverted your PR as well. 

But I'll merge it again, do some testing and if everything works I'll push out a new dev-version for testing.

bbernhard:
I would also add the `--no-log-init` flag to the `useradd` command. The docker [documentation](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user) mentions that there's an unresolved bug that could to disk exhaustion under certain circumstances. Not sure if we are triggering that bug, but I think it shouldn't hurt to add that flag. Any objections from your side?

> Due to an unresolved bug in the Go archive/tar package’s handling of sparse files, attempting to create a user with a significantly large UID inside a Docker container can lead to disk exhaustion because /var/log/faillog in the container layer is filled with NULL (\0) characters. A workaround is to pass the --no-log-init flag to useradd. The Debian/Ubuntu adduser wrapper does not support this flag.

fl42:
> I would also add the --no-log-init flag to the useradd command. The docker documentation mentions that there's an unresolved bug that could to disk exhaustion under certain circumstances. Not sure if we are triggering that bug, but I think it shouldn't hurt to add that flag. Any objections from your side?

I was not aware of this bug. Not sure is we are concerned (`a significantly large UID`), but the flag looks safe anyway.

Thank you for merging
Do you plan to wait few days/weeks to have feedback before merging to latest tag?

bbernhard:
> Do you plan to wait few days/weeks to have feedback before merging to latest tag?

To be honest, I want to get that version out there as fast as possible. Originally I wanted to ship your changes together with the latest signal-cli version (0.7.1), but it turned out that there were some pretty big changes in version 0.7.0 which also requires quite a few changes in the Dockerfile. So, I think it's better to ship your changes first and then work on the integration of signal-cli v0.7.1.

I've just pushed a new docker image to dockerhub which contains your changes. It can be pulled with
`docker pull bbernhard/signal-cli-rest-api:0.31-dev`. If there are no issues with this docker image, this image will then also be used for the `latest` tag. 

I'll later notify all the contributors of this repository..maybe someone wants to help a bit with testing. I would say we wait a few days for feedback and if no objections are raised I'll update the `latest` tag.

PR #43, opened 2020-11-22
JRE Dependency Fix + Port Configuration

### LINKED ISSUE #39: Receiving 400 Response and API Timeout
I have the Docker container up and running, but I'm getting a `400` HTTP response as well as a `{"error":"process killed as timeout reached"}` message in the body.  I haven't tried all endpoints, but `/v1/about` works just fine, but anytime I try to register a number it times out.

Has anyone else run into this?  I'm not seeing much else in logs unfortunately.  I **did** modify the code I'm running to run on port `3000`.  I'm running this on Docker Swarm and port 8080 was already in use.  Are there places that port 8080 is expected that I might be missing?

A way to configure the port at build/runtime would be super helpful if that is what is causing the issue I'm having.

### LINKED ISSUE #42: Latest dev images not running on ARM (release blocker)
As you've might have noticed, the `bbernhard/signal-cli-rest-api:latest` tag on dockerhub hasn't been updated in a while. The reason for that is, that the "cross-built" (with `docker buildx`, see build script [here](https://github.com/bbernhard/signal-cli-rest-api/blob/master/publish.sh)) ARM image which is built on x86-64 doesn't work on the ARM machine. 

If you want to give it a shot, just pull this `bbernhard/signal-cli-rest-api:latest-dev` image from dockerhub and run it on an ARM machine. When you execute a send request with curl, the request fails after 60 seconds with: `"error":"process killed as timeout reached"`. 

The reason for that is, that `signal-cli` inside the docker container blocks forever. This can easily be reproduced by connecting to the running docker container with `docker exec -it <container id> /bin/bash` and then executing `signal-cli`. As you can see, the `signal-cli` binary blocks forever. 

I am still investigating, but I think it has something to do with seccomp and Ubuntu 20.04. see this reports here: 

https://bugs.launchpad.net/ubuntu/+source/libseccomp/+bug/1890332
https://github.com/debuerreotype/docker-debian-artifacts/issues/97

As the `adoptopenjdk/openjdk11` image we are using is based on Ubuntu 20.04 and I see the same issue when doing an `apt-get update` inside the docker container, I guess we could run into this issue. I've tried to use `strace` to see where the `signal-cli` binary is hanging, but unfortunately it's not installed in the `adoptopenjdk/openjdk11` image. But I couldn't pull the package in with apt-get due to the issue above. So I decided to start the container with `--privileged` and now the invocation of `signal-cli` works without problems. 

edit: Okay, I've pretty much confirmed that it's the above issue:

If the docker container is started with 

`docker run -it --entrypoint /bin/bash bbernhard/signal-cli-rest-api:latest-dev`

and you invoke `signal-cli` in there, it blocks forever. 

But if you start the docker container with 

`docker run -it --security-opt seccomp:unconfined --entrypoint /bin/bash bbernhard/signal-cli-rest-api:latest-dev`

`signal-cli` works.

funkybunch:
Not seeing a `CONTRIBUTING.md` file so I'll do my best to be specific.

Main change: Changes the JRE image to a `Bionic`-specific image.  This change is in `Dockerfile` line 42.

This PR also allows the port to be configured through `docker-compose.yml`.  This means images built using `docker-compose` will use the configured port, but images built using `docker build` will still default to port 8080.  The environment variable definitions required to do this were added to `docker-compose.yml`.  Nothing breaks if the environment variables are omitted - the API just runs on port 8080 like it does currently.

A small change is made to `main.go` to detect the `PORT` environment variable and use it to configure Swagger if it exists, but fallback to `8080` if it does not (like if it is built with the Dockerfile).

This PR will resolve #39 and also resolve #42

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,7 +39,7 @@ COPY src/go.sum /tmp/signal-cli-rest-api-src/
 RUN cd /tmp/signal-cli-rest-api-src && swag init && go build
 
 # Start a fresh container for release container
-FROM adoptopenjdk:11-jre-hotspot
+FROM adoptopenjdk:11-jdk-hotspot-bionic
 
 COPY --from=buildcontainer /tmp/signal-cli-rest-api-src/signal-cli-rest-api /usr/bin/signal-cli-rest-api
 COPY --from=buildcontainer /tmp/signal-cli /opt/signal-cli
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,8 @@ version: "3"
 services:
   signal-cli-rest-api:
     build: "."
+    environment:
+      - PORT=8080
     ports:
       - "8080:8080" #map docker port 8080 to host port 8080.
     volumes:
--- a/src/main.go
+++ b/src/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"os"
 
 	"github.com/gin-gonic/gin"
 	log "github.com/sirupsen/logrus"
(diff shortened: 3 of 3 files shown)


bbernhard:
very nice, many thanks for the contribution!

PR #57, opened 2021-01-16
Add healthcheck

x-jokay:
Adds a [healthcheck](https://docs.docker.com/engine/reference/builder/#healthcheck).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,5 +96,6 @@ RUN groupadd -g 1000 signal-api \
 EXPOSE 8080
 
 ENTRYPOINT ["/entrypoint.sh"]
-	
 
+HEALTHCHECK --interval=20s --timeout=10s --retries=3 \
+    CMD curl -f http://localhost:8080/v1/about || exit 1


bbernhard:
Just out of interest: Did you experience any problems with the service not being responsive? I mean it probably won't hurt to have a healthcheck in there, but in case you've experienced some actual issues I would prefer do add some debug logging to track the issue down. :)

x-jokay:
This is not a question of responsiveness but of completeness :wink:

You will be able to see the state using `docker ps` or some monitoring systems.

```sh
CONTAINER ID        IMAGE                                    COMMAND                  CREATED             STATUS                    PORTS                                                              NAMES
3b93a07859d1        bbernhard/signal-cli-rest-api:latest     "/entrypoint.sh"         13 minutes ago      Up 13 minutes (healthy)   8080/tcp                    
```

e.g. Up 13 minutes **(healthy)**.

bbernhard:
> This is not a question of responsiveness but of completeness wink no problems seen so far.
> 
> You will be able to see the state using `docker ps` or some monitoring systems.
> 
> ```shell
> CONTAINER ID        IMAGE                                    COMMAND                  CREATED             STATUS                    PORTS                                                              NAMES
> 3b93a07859d1        bbernhard/signal-cli-rest-api:latest     "/entrypoint.sh"         13 minutes ago      Up 13 minutes (healthy)   8080/tcp                    
> ```
> 
> e.g. Up 13 minutes **(healthy)**.

Oh, didn't know that. Ok, then let's merge the change ;-). Thanks!

x-jokay:
I would even provide a `/health` endpoint for this use case which gets excluded for the logging, so it won't spam the logs.

PR #60, opened 2021-01-16
Switch runtime to JRE

### LINKED ISSUE #59: Use JRE for runtime container
Is there any reason why you use the JDK container instead of the JRE, see [here](https://github.com/bbernhard/signal-cli-rest-api/blob/master/Dockerfile#L75)?

I would switch to [adoptopenjdk:11-jre-hotspot-bionic](https://hub.docker.com/layers/adoptopenjdk/library/adoptopenjdk/11-jre-hotspot-bionic/images/sha256-403072dcaabacb010ba3cddbeaef233de7f083b8e223f87940b006da4905d3f5?context=explore) if possible.

- 11-**jdk**-hotspot-bionic - compressed size: **223.97 MB**
- 11-**jre**-hotspot-bionic - compressed size: **80.3 MB**

x-jokay:
Switches the runtime container from JDK to JRE.

Resolves #59.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,7 +72,7 @@ COPY src/go.sum /tmp/signal-cli-rest-api-src/
 RUN cd /tmp/signal-cli-rest-api-src && swag init && go build
 
 # Start a fresh container for release container
-FROM adoptopenjdk:11-jdk-hotspot-bionic
+FROM adoptopenjdk:11-jre-hotspot-bionic
 
 ARG SIGNAL_CLI_VERSION
 


bbernhard:
Many thanks for the PR!

marvinruder:
@bbernhard Just wanted to suggest the exact same thing (using `17-jre-focal` or even something more fancy like `21-jre-alpine`), was there a reason for switching back to a JDK container in https://github.com/bbernhard/signal-cli-rest-api/commit/5ccf7a9eff2cd577fa2d0d837b57800c6ca5a485#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R115?

marvinruder:
The current images are based on `eclipse-temurin` already, it's just the JDK version.

PR #79, opened 2021-01-28
group messaging fix and more verbose error message

jozef:
Hello,

there are 3 commits in this pull request:

"more verbose/helpful message" → to add some more info when calling qrcodelink. "device_name" parameter is undocumented in Signal Cli REST API documentation yet mandatory.

"-g recipient/group-id needs to stay base64 encoded" → there was base64 decode string and decoded binary passed to "-g" parameter of signal-cli. I've left the check and used the original base64 id string for "-g" flag of signal-cli.

"auto detect group-id recipient" → based on examples using group-id as recipient should work out of the box, but there was/is undocumented "is_group" flag. I've added that unless recipient is a phone number then it's considered to be a group-id.

Best regards
Jozef

### DIFF
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -750,7 +750,7 @@ func (a *Api) GetQrCodeLink(c *gin.Context) {
 	deviceName := c.Query("device_name")
 
 	if deviceName == "" {
-		c.JSON(400, gin.H{"error": "Please provide a name for the device"})
+		c.JSON(400, gin.H{"error": "Please provide a name for the device using ../qrcodelink?device_name=<device_name> parameter"})
 		return
 	}
 
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -175,13 +175,13 @@ func send(c *gin.Context, attachmentTmpDir string, signalCliConfig string, numbe
 			return
 		}
 
-		groupId, err := base64.StdEncoding.DecodeString(recipients[0])
+		_, err := base64.StdEncoding.DecodeString(recipients[0])
 		if err != nil {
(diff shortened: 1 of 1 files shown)


bbernhard:
Thanks for your contribution!

The first commit looks good to me, but I am not sure if I am understanding the other two correctly. 

If you use the `groups` endpoint for listing all the groups, the API returns an `id` as well as an `internal_id`. The `internal_id` is the one that `signal-cli` needs. The `id` (which is the `id` the REST API operates on) is constructed like that: `id = "group." + base64.StdEncoding.EncodeToString([]byte(internalId))`. So the already base64 encoded group id gets base64 encoded again and the `group.` string is prepended - so the group information is now base64 encoded twice. That's why we need to decode it again in the `send` method. 

In retrospect it would have been sufficient to just take the base64 encoded string as it is, without base64 encoding the string again. I guess I've missed the part in the `signal-cli` documentation where it states that the group id is already base64 encoded. But in order to stay backwards compatible and do not break existing setups, I am afraid we have to live with that. 

Regarding your third commit: In order to determine whether the message goes to a group, it is checked whether the recipient starts with the string `group.` (see https://github.com/bbernhard/signal-cli-rest-api/blob/master/src/api/api.go#L554)

jozef:
Oh I see, the pull request is pretty pointless now that you've explained how it works with the group-ids. Sorry for bothering you.

jozef:
Just found out why I needed to patch the api.go to work for me → because I was calling the /v1/send endpoint, while the group-ids with "group." prefix works only with /v2/send.

PR #109, opened 2021-03-13
Update Dockerfile to reduce container size

haberda:
Without the *.tar the tar file is not deleted. This change saves about 20mb of space.


haberda:
I am not sure why the build failed. I cannot build the one in the master repo on my local machine anymore either. Odd

bbernhard:
Good catch, many thanks for the PR!

Regarding the CI: Sometimes the build fails with an out of memory error while building signal-cli with gradle. There are some settings to limit the RAM while building, but I was too lazy so far to implement them.😅. Usually another build with github actions fixes the build.

PR #132, opened 2021-05-16
Add architecture specific lib's for GraalVM

webzit:
Add architecture specific libsignal_jni.so && libzkgroup.so for GraalVM (e.g. Raspberry 64)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,7 @@ RUN cd /tmp/ \
 	&& cd signal-cli-${SIGNAL_CLI_VERSION} \
 	&& git checkout v${SIGNAL_CLI_VERSION} \
 	&& cp /tmp/libzkgroup.so ./lib/src/main/resources/libzkgroup.so \
-        && cp /tmp/libsignal_jni.so ./lib/src/main/resources/libsignal_jni.so \
+	&& cp /tmp/libsignal_jni.so ./lib/src/main/resources/libsignal_jni.so \
 	&& ./gradlew build \
 	&& ./gradlew installDist \
 	&& ./gradlew distTar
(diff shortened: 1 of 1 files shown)


bbernhard:
Thanks for the PR!

Just out of interest: Do you maybe remember which error you got with the latest signal-cli-rest-api docker image and the native mode on arm64? I do not own a arm64 based device, but I tried the arm64 image with QEMU and didn't experience any issues. So just curious which error you got :)

webzit:
{ "error": "WARN App - WARNING: Support for new group V2 is disabled, because the required native library dependency is missing: libzkgroup\nERROR App - Missing required native library dependency: libsignal-client\n" }

After copying the compiled versions of the libraries to ./lib/src/main/resources everything works fine :)

bbernhard:
> { "error": "WARN App - WARNING: Support for new group V2 is disabled, because the required native library dependency is missing: libzkgroup\nERROR App - Missing required native library dependency: libsignal-client\n" }
> 
> After copying the compiled versions of the libraries to ./lib/src/main/resources everything works fine :)

Many thanks for the info! 

Really odd that it only happens with arm64 and the other architectures are fine without that patch. :thinking: 
The change looks pretty harmless to me, so I think it shouldn't break anything for the other architectures. But in order to be sure, I've just kicked off another dev build here: https://github.com/bbernhard/signal-cli-rest-api/actions/runs/864810719

When the job is finished (it usually runs ~1.3hr) there should be a new docker image on dockerhub with the tag `bbernhard/signal-cli-rest-api:0.65-dev` which contains your changes. I'll do some regression testing for armv7 and x86-64 in the next days to make sure that this change doesn't have any side effects for the other architectures. In case it behaves fine, the PR is ready to be merged :tada:. So it will probably take another 1-2 days. Hope that's fine your you.

Thanks a lot for the PR!

bbernhard:
I was finally able to reproduce the issue with QEMU by emulating arm64 on my x86-64 machine and can confirm that your fix also works there. No regressions found on other architectures. 

Thanks again for the fix - it's really great to have the native mode also working on arm64 :tada: 

A new image will be on dockerhub in a few hours.

PR #143, opened 2021-06-12
Added Expiration for Contact

fieldofsheep:
Added Expiration for Contact to Signal CLI.  I haven't much of a clue of how .go works so you might need to re-run the swagger to pull the doc interface.  its listed under v1 update contact and the API appears to work.  User must set the expiration prior to sending a message as previous messages aren't linked to the new timeout.

### DIFF
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -109,6 +109,12 @@ type CreateGroupResponse struct {
 	Id string `json:"id"`
 }
 
+type UpdateContactRequest struct {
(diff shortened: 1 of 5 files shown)


bbernhard (commented):
Many thanks for the PR!

I've added some feedback and improvement suggestions. All the things that I've found are just some minor issues, but I would like to address them already at this point, as breaking the API later is always a bit complicated.

PR #168, opened 2021-10-19
fix code example (tabs in yaml)

lkwg82:


lkwg82:
with asciidoc we could easily include the `docker-compose.yml` and verify in a pipeline

PR #172, opened 2021-10-29
PR to include client expiration

fieldofsheep:
Working on the last PR was too out of date with Merge Conflicts so I bedded in the changes.  Linter must have cleaned some aspects of the file too.  This encompasses https://github.com/bbernhard/signal-cli-rest-api/pull/143

### DIFF
--- a/src/client/client.go
+++ b/src/client/client.go
@@ -733,9 +733,6 @@ func (s *SignalClient) DeleteGroup(number string, groupId string) error {
 }
 
 func (s *SignalClient) UpdateContact(number string, recipient string, name string, expiration_time int) error {
-	println(name)
-	println(number)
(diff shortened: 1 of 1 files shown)


bbernhard (commented):
Sorry, completely forgot to review your changes!

A few things I've noticed:

* I think it would be better to rename `func (a *Api) Contact(c *gin.Context)` to `func (a *Api) UpdateContact(c *gin.Context)` to make it clear that the function updates the contact information.

* As the request only updates the contact info, I think it would be better to use `PUT` instead of `POST`.

* At the moment (almost) all the endpoints use the plural form. So I think it would be better to change the endpoint to `/v1/contacts` instead of `/v1/contact`

* The JSON-RPC implementation is missing.

PR #176, opened 2021-11-01
Reactions Endpoint Proposal

filipre:
Hey, 

I had a look at your reactions branch and I tried to migrate it to the new JSON-RFC standard. Now, signal-cli allows to send reactions to a group and also to remove them so I wanted to build this in as well. What do you think about the proposed API?

For now, I did not write any documentation or even test it (I did not set up Java or Go) but once you are fine with the API design, we can do everything else.

Best!

Update: I changed the optional remove property to a DELETE endpoint

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,9 +1,11 @@
 version: "3"
 services:
   signal-cli-rest-api:
-    image: bbernhard/signal-cli-rest-api:latest
+    #image: bbernhard/signal-cli-rest-api:latest
+    build: .
     environment:
-      - MODE=normal #supported modes: json-rpc, native, normal
+      # - MODE=normal #supported modes: json-rpc, native, normal
+      - MODE=json-rpc #supported modes: json-rpc, native, normal
       #- AUTO_RECEIVE_SCHEDULE=0 22 * * * #enable this parameter on demand (see description below)
     ports:
       - "8080:8080" #map docker port 8080 to host port 8080.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,11 +1,9 @@
 version: "3"
 services:
   signal-cli-rest-api:
-    #image: bbernhard/signal-cli-rest-api:latest
-    build: .
+    image: bbernhard/signal-cli-rest-api:latest
     environment:
-      # - MODE=normal #supported modes: json-rpc, native, normal
-      - MODE=json-rpc #supported modes: json-rpc, native, normal
+      - MODE=normal #supported modes: json-rpc, native, normal
       #- AUTO_RECEIVE_SCHEDULE=0 22 * * * #enable this parameter on demand (see description below)
     ports:
       - "8080:8080" #map docker port 8080 to host port 8080.
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -9,8 +9,8 @@ import (
 
 	"github.com/gabriel-vasile/mimetype"
 	"github.com/gin-gonic/gin"
-	log "github.com/sirupsen/logrus"
 	"github.com/gorilla/websocket"
+	log "github.com/sirupsen/logrus"
 
 	"github.com/bbernhard/signal-cli-rest-api/client"
 	utils "github.com/bbernhard/signal-cli-rest-api/utils"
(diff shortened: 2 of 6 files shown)


bbernhard:
Thanks for the PR!

I definitely like the explicit `DELETE` endpoint more than the `remove` property.

Not sure about the actual endpoint name though. I definitely like the `POST /v1/react`, but I am unsure about the `DELETE /v1/react`. The latter one doesn't sound natural to me. I am thinking whether we should use `POST/DELETE /v1/reactions` instead? What do you think? Another alternative would be to use `/v1/react` for creating and `/v1/reactions` for deleting reactions. But then we would have two different API endpoints to creating/deleting the same resource which is also a bit odd?  

I mean the endpoint in general feels a bit odd and not really restful. In an ideal world, we would probably have something like `POST/DELETE /v1/<number>/messages/<messageid>/reactions`. But as we do not have a unique message id, I guess we don't have any other options than identifying the concrete message via the `recipient`, `timestamp`, `target_author` triple. It's certainly not the nicest way, but I guess we don't have any other options here.

filipre:
According to [this](https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/) an API should use nouns instead of verbs and plural should be used for collections. Therefore, the best name would be `reaction`. It also mentions "resource nesting" but suggest to stop it once you are beyond two levels and there are many APIs that do it using the payload as well, especially for things like a search. Alternatives would be query parameters (worse) or headers (much worse). And, for the typing indicator you also encoded the recipient in the payload.

Just for comparison:
1. `/v1/<number>/reaction`
2. `/v1/<number>/recipient/<recipient>/author/<author>/timestamp/<timestamp>/reaction`
3. `/v1/<number>/<recipient>/<author>/<timestamp>/reaction`

Number one looks the cleanest option to me.

bbernhard:
I would also prefer `1` (especially as all the other endpoints already have the same structure. If I would start all over again, I would probably use a different endpoint structure, but yeah ;-)).

The only thing that still looks a bit "strange" to me is, that we are mixing plural and singular nouns (e.g `/v1/reaction` vs `/v1/groups`). I vaguely remember that I've read somewhere that mixing plural and singular nouns isn't best practice. On the other hand, we already have `/v1/typing-indicator`, so we are already mixing :/

> Should I create the rest or do you want to take over?

Both is fine for me. 

> I struggle a bit to set up a dev environment (how do you compile go? Do yo use Docker for testing, etc.) but if you tell me how you do it, I can also try it out.

Most of the time I am just building the docker image with `docker-compose build`, followed by `docker-compose up` to start the container. If you have `go` installed on your dev machine, you can compile the source by running `go build` inside the `src` folder. But in order to test the changes you then again need a working test environment (with `signal-cli`, `supervisord`, etc). Setting that up manually is pretty time consuming (and error prone), that's why I am using `docker-compose build` most of the time.

filipre:
Great. I agree that being consistent is also important. Thus, I changed it `reactions`. I just implemented everything else and also tested it. I was able to add and remove reactions. For removing reactions, I made the emoji optional because it does not matter what you send there as long as signal-cli gets something. Could you review the PR? 

Also, for some reason, I am getting this error now
```
signal-cli-rest-api_1  | + set -e
signal-cli-rest-api_1  | + [ -z /home/.local/share/signal-cli ]
signal-cli-rest-api_1  | + chown 1000:1000 -R /home/.local/share/signal-cli
signal-cli-rest-api_1  | + cat
signal-cli-rest-api_1  | + cap_prefix=-cap_
signal-cli-rest-api_1  | + cat /proc/sys/kernel/cap_last_cap
signal-cli-rest-api_1  | + seq -s ,-cap_ 0 40
signal-cli-rest-api_1  | + caps=-cap_0,-cap_1,-cap_2,-cap_3,-cap_4,-cap_5,-cap_6,-cap_7,-cap_8,-cap_9,-cap_10,-cap_11,-cap_12,-cap_13,-cap_14,-cap_15,-cap_16,-cap_17,-cap_18,-cap_19,-cap_20,-cap_21,-cap_22,-cap_23,-cap_24,-cap_25,-cap_26,-cap_27,-cap_28,-cap_29,-cap_30,-cap_31,-cap_32,-cap_33,-cap_34,-cap_35,-cap_36,-cap_37,-cap_38,-cap_39,-cap_40
signal-cli-rest-api_1  | + [ json-rpc = json-rpc ]
signal-cli-rest-api_1  | + /usr/bin/jsonrpc2-helper
signal-cli-rest-api_1  | /entrypoint.sh: 24: /usr/bin/jsonrpc2-helper: Exec format error
```
Could you test it if you also get this error?

bbernhard:
> Could you test it if you also get this error?

Everything works fine for me here. 

The diff looks also good. I've kicked of a CI build (which builds the docker image for all the architectures) and if that one runs through too, I think we can merge it :)

filipre:
Awesome. Looks good  then :) Maybe I need to rebuild it for whatever reasons? I also notice that there is a `swagger.json`, a `swagger.yaml` and a `docs.go` file that all describe the same thing. Is that really necessary? I think it would be better if there is only one file at some point.

bbernhard:
> Maybe I need to rebuild it for whatever reasons?

Yeah, I guess so.

> Is that really necessary?
The swagger documentation is auto generated by [swag](https://github.com/swaggo/swag). So, you don't need to touch any of those files manually, but just add the appropriate sourcecode annotations and let `swag` do the work.

The build looks good :+1:. I'll merge it now, thanks for the PR!

filipre:
Thanks for merging! Could you release a dev version as well?

bbernhard:
> Thanks for merging! Could you release a dev version as well?

sure. It's here: `bbernhard/signal-cli-rest-api:0.82-dev`

PR #179, opened 2021-11-10
Improve HOMEASSISTANT.md

akloeckner:
This addresses https://github.com/home-assistant/home-assistant.io/issues/20206

It also cleans up the doc a bit using lists and headings and block code.


akloeckner:
An open point would be to add the troubleshooting of joining v2 groups. But I have not figured out the clean way to do this, so I wouldn't want to give bad hints.

bbernhard:
Many thanks for the great documentation improvement - very much appreciated!

Issue #201, opened 2022-02-02
Native Mode causes 'RuntimeException: Field storyContext_ for org.whispersystems.signalservice.internal.push.SignalServiceProtos$DataMessage not found'


bentolor:
### The problem

Thanks for this helpful container for Homeassistant Signal Integration!

To circumvent high latency for sending messages, I tried to use native mode. I stumbled over two issues

## Issue 1: Documentation
The README is talking about modes, but I found nowhere a clue what the default behavior is nor on how to understand which mode is active or how to control the mode.

Eventually I started to search the repo for `/usr/bin/signal-cli-native` and reading the Go code to understand how it is working. 

A note hinting that you need to set/pass a `MODE` on container creation as well as a hint on startup on the current mode (`Started Signal Messenger REST API is working in XXX mode`) would be great.

## Issue 2: Native mode not working

Setting `MODE=native` leads to a non-working installation returning:

[...]

PR #204, opened 2022-02-05
Describe mode usage and provide quickstart tutorial

bentolor:
Partially addresses #201


bbernhard:
Many thanks for the PR!

PR #206, opened 2022-02-05
build(Dockerfile): reduce build noise to a minimum

bentolor:
This depends on #205

This PR tries to reduce the ouput noise during image build.


bbernhard:
Many thanks!

PR #208, opened 2022-02-05
Added support for missing send arguments

robin-thoni:
Added support for missing send arguments:
- Mentions
- Quote timestamp
- Quote author
- Quote message
- Quote mentions

This is tested locally with `normal` and `json-rpc`.

~I'm not sure if a `v3` is required. If you think not, I can merge it back to `v2`.~ Edit: it is, otherwise clients have no way to know if mentions and quotes are supported or not (and would simply be lost)

WARNING: this my first ever experience w/ Go, please pay extra attention while reviewing.

### DIFF
--- a/src/client/client.go
+++ b/src/client/client.go
@@ -120,6 +120,7 @@ type About struct {
 	BuildNr              int      `json:"build"`
 	Mode                 string   `json:"mode"`
 	Version              string   `json:"version"`
+	Capabilities         map[string][]string    `json:"capabilities"`
 }
 
 type SearchResultEntry struct {
@@ -283,7 +284,8 @@ func (s *SignalClient) Init() error {
 }
 
 func (s *SignalClient) send(number string, message string,
-	recipients []string, base64Attachments []string, isGroup bool) (*SendResponse, error) {
+	recipients []string, base64Attachments []string, isGroup bool, mentions []string,
+	quote_timestamp *int64, quote_author *string, quote_message *string, quote_mentions []string) (*SendResponse, error) {
 
 	var resp SendResponse
 
@@ -324,10 +326,15 @@ func (s *SignalClient) send(number string, message string,
 		}
 
 		type Request struct {
-			Recipients  []string `json:"recipient,omitempty"`
-			Message     string   `json:"message"`
-			GroupId     string   `json:"group-id,omitempty"`
-			Attachments []string `json:"attachment,omitempty"`
+			Recipients  		[]string `json:"recipient,omitempty"`
+			Message     		string   `json:"message"`
+			GroupId     		string   `json:"group-id,omitempty"`
+			Attachments 		[]string `json:"attachment,omitempty"`
+			Mentions          	[]string `json:"mentions,omitempty"`
+			QuoteTimestamp  	*int64   `json:"quote-timestamp,omitempty"`
+			QuoteAuthor       	*string  `json:"quote-author,omitempty"`
+			QuoteMessage      	*string  `json:"quote-message,omitempty"`
+			QuoteMentions     	[]string `json:"quote-mentions,omitempty"`
 		}
 
 		request := Request{Message: message}
@@ -339,6 +346,11 @@ func (s *SignalClient) send(number string, message string,
 		for _, attachmentEntry := range attachmentEntries {
 			request.Attachments = append(request.Attachments, attachmentEntry.toDataForSignal())
 		}
+		request.Mentions = mentions
+		request.QuoteTimestamp = quote_timestamp
+		request.QuoteAuthor = quote_author
+		request.QuoteMessage = quote_message
+		request.QuoteMentions = quote_mentions
 
 		rawData, err := jsonRpc2Client.getRaw("send", request)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


bbernhard:
Many thanks for the PR!

I haven't looked in detail over your PR, but I think you were just adding new members, right? In that case we should be API compatible and there shouldn't be any need to create a `v3`. :)

robin-thoni:
The only thing I'm worried about is that a client would then have no way to know if mention and quote related arguments will be handled or simply ignored

robin-thoni:
>  I am afraid that we will end up with API endpoints in a lot of different versions.
Yeah, that's definitely a valid concern.

> but can we maybe work around that problem by introducing a dedicated API endpoint for that?
Then, we might end up w/ the above situation: what's gonna happen when Signal or signal-cli add something new? A simple example would be the "View once message" feature. I don't think signal-cli supports that already, but when it will, we're going to be back to the same situation again...

Maybe we could simply add a `capabilities` endpoint or key in `about`? Then we can just push some values in a list, like `{"v2/send": ["quotes", "mentions"]}`. That way, we keep the existing `v2` endpoint, but we can extend it to infinity and beyond. What do you think?

bbernhard:
Sorry for the late reply!

A `capabilities` list inside the `about` endpoint sounds good. I really like that 👍

robin-thoni:
Ok, perfect, I'll implement that on the following days (was out of town).

robin-thoni:
Ok, I'm back to it. Sorry for the delay. I pushed the changes we talked about:
- No `v3` anymore
- `capabilities` field in `about` endpoint

Again, everything has been tested locally. Here are a few examples sent w/ the new options (mentions, quote, quote text, quote mentions):
Screenshot from 2022-04-03 19-12-55

robin-thoni:
Well, actually, I was thinking about splitting mentions into a nested object, like
```
{
  "start": 0,
  "length": 5,
  "author": "+14242424242"
}
```
rather than a simple string `0:5:+14242424242`, as expected by `signal-cli`. What do you think?

bbernhard:
Thanks for the rework - looks really nice!

> What do you think?

Would definitely prefer the nested object over the string.

But personally I still find it a bit uncomfortable to use. I mean, I get why signal-cli splitted the whole thing up in two different parameters, but from a user experience perspective I find it suboptimal. 

I was thinking whether we can embed the mention directly into the message string. So e.g

```curl -X POST -H "Content-Type: application/json" -d '{"message": "Hi @+999999!", "mention": true, "number": "+1234567", "recipients": ["group.xxxxxxx"]}' 'http://127.0.0.1:8080/v2/send```

In case `mention` is set to `true` in the payload, we parse the `message` string and look for `@` characters followed by a number in international phone number format (i.e a number starting with `+`). If we find those, we transform the payload into something that signal-cli understands. So in the above example it would e.g be: `-m "Hi X!" --mention "3:1:+999999"`

What do you think?

**Comment by robin-thoni** [comment]
Oops, sorry, I missed your answer.

I personally prefer the metadata idea, rather than parsing a text message, especially when it's about an API that could be used by monitoring systems, and thus, could contain any arbitrary data. Because of that, I'd suggest to stick w/ the mention object. Any objection?

**Comment by mattwr18** [comment]
Is there anything we can do to help get this merged in @bbernhard @robin-thoni? We would very much like to be able to quote messages. Thanks again for the very useful tool.

**Comment by robin-thoni** [comment]
I'll try to wrap it up this week

**Comment by robin-thoni** [comment]
Sorry, I didn't had time. I'm now out of town. Hopefully, I'll have time on the 7th November week

**Comment by mattwr18** [comment]
> Sorry, I didn't had time. I'm now out of town. Hopefully, I'll have time on the 7th November week

Thank you for the update. Appreciate the work you've put in so far and looking forward to being able to use this feature :star_struck:

File: src/client/client.go [hunk 1]
```diff
@@ -282,8 +289,13 @@ func (s *SignalClient) Init() error {
...
 }
 
+func (s *MessageMention) toString() string {
+    return fmt.Sprintf("%d:%d:%s", s.Start, s.Length, s.Author)
+}
+
 func (s *SignalClient) send(number string, message string,
-	recipients []string, base64Attachments []string, isGroup bool) (*SendResponse, error) {
+	recipients []string, base64Attachments []string, isGroup bool, mentions []MessageMention,
+	quote_timestamp *int64, quote_author *string, quote_message *string, quote_mentions []MessageMention) (*SendResponse, error) {
```
bbernhard, line 298: Thanks for the changes! Just a small thing, but could you please make the variable names camelCased. Many thanks! Apart from that it looks fine.

robin-thoni: Well caught! Too many copy/paste...

robin-thoni:
Hi @bbernhard sorry to bother you, this is just to make sure you got notified for the fix above, following your review.

Thanks

bbernhard:
Thanks for the ping @robin-thoni - I somehow missed the github notification. It is now merged. Many thanks!

robin-thoni:
@mattwr18 you might be interested in this: bbernhard/pysignalclirestapi#10

mattwr18:
@robin-thoni interesting. I'll have to look into that further. For the time being, this is super helpful for us, I guess. Still need to wait for the release and implement it on our end, but super excited about this possibility. Thank you again for your contribution and @bbernhard for reviewing and getting this merged on top of maintaining this API.

PR #209, opened 2022-02-06
Update docs.go

KY-7:
Missing attributes definitions


bbernhard:
Thanks for the PR!

We are using [swag](https://github.com/swaggo/swag) to generate the swagger documentation. So, instead of directly changing the `docs.go` file, the REST API code is annotated and `swag` will then be used to generate the documentation.

PR #289, opened 2022-10-08
Prepare upgrade to signal-cli 0.11.3 tpo solve current issues with Signal

bentolor:
Due to some major issues with the older signal-cli due to API deprecation on server-side, I'm trying to built an up-to-date version of signal-cli-rest-api. (Issues: Unable to send/receive message due to DeprecatedExceptions, HTTP 499, etc. Upgrading signal-cli causes unreadable config file, etc.).

I [already contributed to fixed the native build](https://github.com/AsamK/signal-cli/pull/1037) of the current `signal-cli`. On integration it now seems that this requires libsignal 0.20.0.

I tried to build that as well, but fail for the other platforms. Obviously `cross` or `cargo`  are  picking the wrong linker, hence the linkage of the cross-plattform binaries fails. Not sure if this is running for you already, @bbernhard.

So what IMHO remains to provide a hotfix release is to build the libsignal libs and updating the version.


bbernhard:
Many thanks!

I gave up on cross compiling and now build the libraries for armv7 directly on the Pi and for arm64 emulated with QEMU.

PR #353, opened 2023-04-05
Add stickers support in API and client

Matovidlo:
Currnet go API does not support stickers as java cmd line tool / jsonRPC supports it.
Extend method of send v2 to support stickers usage.


bbernhard:
Sorry for the late response!

The PR looks good - many thanks for implementing the feature!

PR #392, opened 2023-08-13
Update HOMEASSISTANT.md

developersteve:
Added a trouble shooting note for linking an account to the api


bbernhard:
Many thanks! :)

PR #399, opened 2023-08-24
Update README.md

developersteve:
added details to the README about the port env var being available to be set


bbernhard:
Thanks a lot - very much appreciated!

Issue #365, opened 2023-05-12
Add support of extra flags for `receive` command


xSAVIKx:
### Feature Request

The `receive` endpoint now only supports the `timeout` argument, but the CLI also provides the following configs:

```text
usage: signal-cli receive [-h] [-t TIMEOUT] [--max-messages MAX_MESSAGES] [--ignore-attachments] [--ignore-stories] [--send-read-receipts]

named arguments:
  -h, --help             show this help message and exit
  -t TIMEOUT, --timeout TIMEOUT
                         Number of seconds to wait for new messages (negative values disable timeout)
  --max-messages MAX_MESSAGES
                         Maximum number of messages to receive, before returning.
  --ignore-attachments   Don’t download attachments of received messages.
  --ignore-stories       Don’t receive story messages from the server.
  --send-read-receipts   Send read receipts for all incoming data messages (in addition to the default delivery receipts)
```

It would be great to add support for the other flags as well. IMO `--ignore` flags are especially helpful with the auto receive feature.

k-bx:
When my service which uses REST API isn't up for a long time, and then I try to receive messages, I get `https://github.com/bbernhard/signal-cli-rest-api/issues/365` and messages are lost. I would hope `--max-messages` analog would help.

bbernhard:
> When my service which uses REST API isn't up for a long time, and then I try to receive messages, I get `{"error":"process killed as timeout reached"}` and messages are lost. I would hope `--max-messages` analog would help.

Currently, `--max-messages` isn't exposed via the REST API. But specifying a long timeout (e.q: `GET /v1/receive?timeout=60`) should work too.

PR #401, opened 2023-08-31
Accept max_messages param

k-bx:
Also upgrades signal cli to 0.12.1 since it failed to build previously (couldn't find signal-cli-native of 0.12.0).

Contributes towards https://github.com/bbernhard/signal-cli-rest-api/issues/365


k-bx:
One thing to note on this. This worked on my Docker macOS arm64, but failed on prod Ubuntu x86_64 machine with:

```
WARN  Manager - Failed to call libsignal-client: Can't load library: /tmp/1243969752656717
```

I wanted to try out newer LIBSIGNAL_CLIENT_VERSION=0.30.0 but couldn't find where I can get the files for ext/libraries/libsignal-client/v0.31.0.

Possibly relevant https://github.com/AsamK/signal-cli/issues/1305

bbernhard:
Many thanks for the PR!

I'll wait until the CI is finished, but to me it looks good!

Regarding your issues: Which mode did you use on your x86-64 machine? the native mode?

regarding libsignal-client v0.31.0: I always try to stay at the same version that signal-cli comes packaged with. Newer versions might work, but I think it is the safest approach to use the same version as the pre-packaged signal-cli :)

k-bx:
You're welcome!

> Regarding your issues: Which mode did you use on your x86-64 machine? the native mode?

Native mode

> regarding libsignal-client v0.31.0: I always try to stay at the same version that signal-cli comes packaged with. Newer versions might work, but I think it is the safest approach to use the same version as the pre-packaged signal-cli :)

Gotcha. Where do you see which version it's packaged with?

k-bx:
One thing I forgot to mention, I would very much appreciate if you'd ping me once this is on Docker Hub. I think it might be something with my image being built differently so I'd love to try out the "official" build.

bbernhard:
> Native mode

ok, then it could indeed be GraalVM related.

> Gotcha. Where do you see which version it's packaged with?

It is mentioned in the signal-cli release [changelog](https://github.com/AsamK/signal-cli/blob/master/CHANGELOG.md) and verified again in the [dockerfile](https://github.com/bbernhard/signal-cli-rest-api/blob/master/Dockerfile#L108).

> One thing I forgot to mention, I would very much appreciate if you'd ping me once this is on Docker Hub. I think it might be something with my image being built differently so I'd love to try out the "official" build.

I'll try, if I do not forget. You can btw. also watch this github repository for a new release (see https://github.blog/changelog/2018-11-27-watch-releases/). I'll always create a new github release when a new release is ready and the image is pushed to dockerhub. But maybe I can get docker image up for beta testing sooner, in that case I'll let you know :)

k-bx:
I’m watching all commits and releases so should be good :)

mattwr18:
I was just creating a ticket for this issue. We have been testing out the `0.68-pre` release and ran into this issue in native mode.

PR #405, opened 2023-09-06
Add the ability to refresh a group.

greysonp:
Hi there! I ran into a few issues where my bot's view of group membership would start to fall out of sync with reality. Adding/removing the bot from the group didn't fix things. I was able to find that there was a signal-cli command to re-sync the group state [here](https://github.com/AsamK/signal-cli/issues/1095#issuecomment-1309019924), but there didn't appear to be a way to trigger that from signal-cli-rest-api.

So, I just added a new endpoint `POST /v1/groups/{number}/{groupid}/refresh` that calls the appropriate refresh command. Can confirm that it works and helped solve some long-standing issues with my bots :)

Thanks for building and maintaining this project! It's been incredibly useful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 signal-cli-config
 src/main
 src/signal-cli-rest-api
+.idea/
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -1229,17 +1229,17 @@ func (a *Api) QuitGroup(c *gin.Context) {
 	c.Status(http.StatusNoContent)
 }
 
-// @Summary Refresh the state of a Signal Group.
+// @Summary Update the state of a Signal Group.
 // @Tags Groups
-// @Description Refresh the state of a Signal Group.
+// @Description Update the state of a Signal Group.
 // @Accept  json
 // @Produce  json
 // @Success 204 {string} OK
 // @Failure 400 {object} Error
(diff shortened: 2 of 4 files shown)


bbernhard:
Many thanks for the PR - very much appreciated!

The PR looks good, I just would have one minor suggestion: Could you maybe make it a `PUT` request with the following URL: `/v1/groups/{number}/{groupid}`. The reason why I would prefer a PUT request, instead of a dedicated `/refresh` endpoint is, that I am hoping that we can kill two birds with one stone that way.

Currently, we are missing the functionality to change the name of the group and its description (which would fit a PUT request perfectly imho). So, if the `refresh` gets implemented as `PUT /v1/groups/{number}/{groupid}` the endpoint can later be extended with some optional parameters to change the name of the group and its description. And if one just wants to refresh the group, the endpoint can be called with no parameters.

Another reason, why I am a bit hesitant to expose the refresh as a dedicated `/refresh` endpoint is, that to me, it sounds like a bit of a bug/workaround for some limitation in the signal-cli implementation, which might get fixed in the future.

What do you think? Would that make sense for you?

greysonp:
Makes sense! I pushed the change.

bbernhard:
Thanks a lot!

PR #432, opened 2023-10-29
Allow updating group name

kxait:
`signal-cli` allows for updating the group name along with description and avatar
https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc#updategroup

this PR adds that functionality to signal-cli-rest-api


bbernhard:
Looks good - many thanks!

PR #445, opened 2023-11-11
Link device in JSON-RPC mode

### LINKED ISSUE #433: Link device in JSON-RPC mode
### Feature Request

Hi, there was an update to signal-cli regarding linking devices in JSON-RPC mode, it would be really nice to have this support in this project, as for e.g. my application I need to be able to link devices on the fly.

Issue: https://github.com/AsamK/signal-cli/issues/1319
Commit: https://github.com/AsamK/signal-cli/commit/a7a5947a1bc468cbf202c603ed14fe40b14c73b0

It requires the json-rpc client to be connected without a `-u` parameter (in multi-account mode), and for there to be 2 requests sent (`startLink` which gets the QR code and `finishLink` which waits for the device to be registered). It seems this would require big changes in the way we handle the JSON-RPC connections

1. Do not assume every json-rpc connection needs a phone number associated with it
2. Add a multi-account mkfifo to signal-cli
   - do not crash if we didn't find any registered numbers at the beginning?
3. Send a request **after** we return the QR code and wait for it to finish or wait for a timeout
4. Regenerate the `jsonrpc2.yml` config after the linking is finished and reload it

So far, I have got it to actually link the number in the background and add it to `data/`, but I think actually reloading `jsonrpc2.yml` would require some big changes and assistance/guidance from @bbernhard (let me know if I'm on the right path)

My branch: https://github.com/bbernhard/signal-cli-rest-api/commit/c7f22838904b7ea33d539dfcbce4a20d0fea0020

```
[GIN] 2023/10/29 - 10:47:31 | 200 |  1.531793584s |      172.17.0.1 | GET      "/v1/qrcodelink?device_name=test"
2023/10/29 10:47:46 Linking device result: {"number":"+48123456789"}
```

kxait:
This PR adds the link-device functionality previously available in `native` to `json-rpc`. As signal-cli requires there to be a second request `finishLink` sent to actually link the device, this runs in a goroutine after the qr code is sent. The goroutine either finishes successfully, or dies after 1 minute due to signal-cli sending a failure message. (the result is available through debug logs).

To accomplish this, the `receivedMessages` chan had to be modified to actually be a map from "message id" (generated by us) to chan, indexed by the message receiver goroutine when a message is received and deleted after the result is parsed. 
This is because when there was an ongoing `finishLink` "job" running, the chan listener would eat all the other responses, so the service locked up while waiting for messages.

Closes #433

### DIFF
--- a/src/client/client.go
+++ b/src/client/client.go
@@ -1103,6 +1103,7 @@ func (s *SignalClient) GetQrCodeLink(deviceName string, qrCodeVersion int) ([]by
 				return
 			}
 			log.Debug("Linking device result: ", result)
+			s.signalCliApiConfig.Load(s.signalCliApiConfigPath)
 		})()
 
 		return png, nil
--- a/src/client/client.go
+++ b/src/client/client.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	log "github.com/sirupsen/logrus"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


bbernhard:
Looks good - thanks a lot!

jamesoncollins:
Could something like this work for v1/register as well?

bbernhard:
> Could something like this work for v1/register as well?

What do you mean exactly?

PR #447, opened 2023-11-12
Get list of accounts

kxait:
Adds `GET /v1/accounts` which returns an array of phone numbers (accounts) available to use.
[The main signal-cli manpage](https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc) does not reference this functionality, but [the manpage for the dbus API](https://github.com/AsamK/signal-cli/blob/master/man/signal-cli-dbus.5.adoc#signalcontrol-interface) does:

```
listAccounts() → accountList<as>
accountList : Array of all attached accounts in DBus object path form
```

Tested with both json-rpc and native mode, confirmed to show linked accounts, I sadly can't test it with number registrations, so it would be nice if somebody could confirm it works!

Also, I noticed that when running `swag i` on the latest version of swaggo, there are **a lot** of additional changes to the swagger spec. @bbernhard what version of swaggo are you using?

### DIFF
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -894,6 +894,23 @@ func (a *Api) GetQrCodeLink(c *gin.Context) {
 	c.Data(200, "image/png", png)
 }
 
+// @Summary List all accounts
+// @Tags Accounts
+// @Description Lists all of the accounts linked or registered
+// @Produce json
+// @Success 200 {object} []string
+// @Failure 400 {object} Error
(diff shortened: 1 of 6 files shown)


bbernhard:
Many thanks!

regarding swag: I am currently using version `1.6.7` which is already pretty outdated. So I guess it makes sense to update my local version of swag :sweat_smile:

PR #450, opened 2023-11-16
Use JRE for runtime container

### LINKED ISSUE #449: Use JRE for runtime container
Is there any reason why you use the JDK container instead of the JRE? This was done before (see #59, #60), but reverted in https://github.com/bbernhard/signal-cli-rest-api/commit/5ccf7a9eff2cd577fa2d0d837b57800c6ca5a485 while switching to Java 17.

I would switch to [`21-jre-jammy`](https://hub.docker.com/_/eclipse-temurin/tags?name=21-jre-jammy) or even [`21-jre-alpine`](https://hub.docker.com/_/eclipse-temurin/tags?name=21-jre-alpine) if possible (Alpine images are not available for `arm/v7` architecture, only v8), reducing the image size by approx. 100 MB.

marvinruder:
Resolves #449.


bbernhard:
Thanks!

PR #474, opened 2024-02-05
Pass JAVA_OPTS through to signal-cli in json-rpc mode

### LINKED ISSUE #473: JAVA_OPTS are ignored when using json-rpc mode
### The problem

To get the signal-cli to work behind a proxy you can using the JAVA_OPTS environment variable to configure them like so:
```
JAVA_OPTS=-Dhttp.proxyHost=proxy.server -Dhttp.proxyPort=3128 -Dhttps.proxyHost=proxy.server -Dhttps.proxyPort=3128
```
However, this fails in json-rpc mode. (It also fails in native mode but I don't know how to fix that.)

It happens because in java-rpc mode signal-cli is run under supervisor as a service, which clears the environment.

The fix is to copy the variable through. Will post patch shortly.

### Are you using the latest released version?

- [X] Yes

### Have you read the troubleshooting page?

- [X] Yes

### What type of installation are you running?

signal-cli-rest-api Docker Container

### In which mode are you using the docker container?

JSON-RPC Mode

### What's the architecture of your host system?

x86-64

### Additional information

_No response_

kleptog:
Fixes #473

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -25,6 +25,9 @@ caps="$cap_prefix$(seq -s ",$cap_prefix" 0 $(cat /proc/sys/kernel/cap_last_cap))
 if [ "$MODE" = "json-rpc" ]
 then
 /usr/bin/jsonrpc2-helper
+if [ -n "$JAVA_OPTS" ] ; then
+    echo "export JAVA_OPTS='$JAVA_OPTS'" >> /etc/default/supervisor
+fi
 service supervisor start
 supervisorctl start all
 fi


bbernhard:
Thanks for the PR - very much appreciated!

Just a small thing: Would it be possible to use `>>` instead of `>`, so that the changes are appended to the file instead of the whole file getting replaced? Right now, it doesn't matter much, as we do not have anything in there, except your `JAVA_OPTS`. But maybe at a later point someone writes something to the file as well. And then your `>` would overwrite those changes again.

What do you think? :)

kleptog:
Fair enough. When writing it I was going either way, but I don't mind.

Update pushed

bbernhard:
Many thanks for the contribution!

Issue #490, opened 2024-02-23
Enable unprivileged start 


fluktuid:
### Feature Request
As an administrator I want to be be able to start this container unprivileged.

### Current situation
the `entrypoint.sh` script is
- trying to chown the SIGNAL_CLI_CONFIG_DIR
- uses exec setpriv …

this causes issues issues when run without privileges

### solution
check wether `uid` and `gid` are also the target `SIGNAL_CLI_UID` and `SIGNAL_CLI_GID`.
If so, just start `signal-cli-rest-api`

poggenpower:
I have created a new approach for a non-root container. In PR https://github.com/bbernhard/signal-cli-rest-api/pull/798 you can find a link to an image for texting. Any feedback is welcome.

PR #491, opened 2024-02-23
enable rootless start

fluktuid:
fixes bbernhard/signal-cli-rest-api#490


bbernhard (changes requested):
Personally I would prefer to wrap the relevant code parts (usermod, groupmod, etc) in the if block, instead of duplicating the code. Duplicating the code makes it harder to maintain and code can easily be lost (e.g in you code part the ` WARNING` message isn't set in the `.bashrc` file and the `HOST_IP` is also not set)

okrammer:
Is there a chance that we get this fixed and merged?

PR #492, opened 2024-02-24
Add sentry lib

KyryloKrasiuk:
Proposal to add Sentry to the project.

Sometimes there are errors in the logs and it is necessary to somehow monitor them.


bbernhard:
Is there a specific reason why you prefer Sentry over other log monitoring solutions that don't require code changes? I've used Sentry in the past and I also quite liked it, but I am not sure if it is a good idea to add that kind of monitoring to a software that is particular privacy focused.

I know, that one must configure a SENTRY_DSN before the application sends data to the Sentry Server, but I imagine that this will result in some questions from users that do not know Sentry well enough and are afraid about their privacy (which I can totally understand).

KyryloKrasiuk:
This is the first option I thought of.
In those cases where I don't have access to the container (SSH) and can't install additional services to monitor logs, this seemed like a good option.

Maybe you have recommendations on how to better collect logs and errors? Because the other option is to make my own Dockerfile based on yours.

poggenpower:
I agree with @bbernhard , this kind of proprietary 3rd party dependencies should not be hard coded into the signal-cli-rest-api. 
@KyryloKrasiuk if you really need sentry integration, patch sentry into signal-cli-rest-api  and build it on your own. From the changes required this should work pretty reliable and this should be easy to automate that you get your preferred logging solution in. 
There are tons of solutions to monitor containers. At the end you can use the signal-cli-rest-api image as base image and add all software you need to do your favorite logging. For containers is it very common to log to stdout, which is collected by the container runtime. It seems https://sematext.com/guides/docker-logs/ is a good starting point if you use docker.

PR #525, opened 2024-04-26
Add a link to signal-cli-to-file

jneidel:
I wrote a utility script over at [signal-cli-to-file](https://github.com/jneidel/signal-cli-to-file) that receives incoming messages from signal-cli-rest-api and save them as files.

It's required for my basic use case for signal-cli:

> My use case is to write notes on my phone, take pictures or record audios that then just show up in my note-taking system. All very convieniently through signal.

Linking to it here so other people don't have to recreate it :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,20 +104,26 @@ services:
       - "./signal-cli-config:/home/.local/share/signal-cli" #map "signal-cli-config" folder on host system into docker container. the folder contains the password and cryptographic keys when a new number is registered
 ```
 
-## Documentation
+## Documentation & Usage
 
 ### API Reference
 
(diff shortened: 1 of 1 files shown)


bbernhard:
Thanks a lot for the PR - very much appreciated!

We already have a table with [third party libraries/clients](https://github.com/bbernhard/signal-cli-rest-api?tab=readme-ov-file#clients--libraries). I guess we could extend that table and not only list clients and libraries but also scripts. And maybe add another optional column with a short description what that component does. :) 

But I am also perfectly fine with the PR as it is right now - if you prefer it that way. Just wanted to raise that in case you've missed it :) Let me know what you think.

bbernhard:
Thanks a lot!

PR #526, opened 2024-04-27
Update README.md

vorwd:
aligned initial docker example with the docker example starting at line 73.  

/home/user/signal-api:/home/.local/share/signal-cli

Having two different file paths (naming signal-api vs signal-cli) was confusing me when I was setting this up and thought it could impact others as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ $ mkdir $HOME/.local/share/signal-cli
 
 ```bash
 $ sudo docker run -d --name signal-api --restart=always -p 8080:8080 \
-      -v $HOME/.local/share/signal-cli:/home/.local/share/signal-cli \
+      -v $HOME/.local/share/signal-api:/home/.local/share/signal-cli \
       -e 'MODE=native' bbernhard/signal-cli-rest-api
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,7 +30,7 @@ $ mkdir $HOME/.local/share/signal-cli
 
 ```bash
 $ sudo docker run -d --name signal-api --restart=always -p 8080:8080 \
-      -v $HOME/.local/share/signal-cli:/home/.local/share/signal-cli \
+      -v $HOME/.local/share/signal-api:/home/.local/share/signal-cli \
       -e 'MODE=native' bbernhard/signal-cli-rest-api
 ```
```
bbernhard, line 35: Are you sure that this worked for you out of the box? :thinking: 

I would assume that you would also need to adapt the line `mkdir $HOME/.local/share/signal-cli` above then. Otherwise you create a folder with the name `signal-cli` on your host system, but bind-mount the `signal-api` folder (which most probably doesn't exist).

vorwd: Yes, you're right.  I missed that bit, as I always create my container folders based on the container name, which is signal-api.  I've updated my pull request.

bbernhard:
Thanks!

PR #524, opened 2024-04-26
Add support for read receipts on /receive

jneidel:
Adds support for [sending read receipts](https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc#receive) on receive.

Closes #431

Issue #431, opened 2023-10-29
Read receipt for messages possible?


Kuhkaefer:
### Feature Request

In order for the user to know that the bot successfully read the message, it would be helpful to see the white check marks as a confirmation.

Would that be possible?

kxait:
hey, you get that from `/v1/receive`

Kuhkaefer:
@kxait sorry, could you elaborate please? I am using 'v1/receive' but the white check marks are not shown. And in the docs I don't see an option to activate read confirmation.

jneidel:
You only get the "delivered to" status from `/v1/receive`.

The underlying cli has a `--send-read-receipts` that is not available through the endpoint.

PR #534, opened 2024-05-11
Add endpoint support for individual read receipts

brettp:
Added `POST /v1/receipt/{number}` endpoint to support read receipts for individual messages, as opposed to all received as currently exists with `/receive`. Complements functionality raised in #431 and #524.

Accepts a payload like:

```
{
  "receipt_type": "read",
  "recipient": "+12345678900",
  "timestamp": 1715386986897
}
```


bbernhard:
Many thanks!

PR #584, opened 2024-09-12
Add missing groupid parameter

derpferd:
The `groupid` parameter was missing from multiple paths.


derpferd:
There are a bunch of unrelated changes in the generated files. I ran `swag init` using version 1.8.10 .

bbernhard:
many thanks!

PR #601, opened 2024-10-14
Add feature to join group using invite link

jailson-dias:
I have added an endpoint to join a group using an invite link. I have created the POST /v1/groups/:number/join_by_invite_link endpoint that receives the invite_link in the body and joins the group using the joinGroup command in signal-cli.


vordenken:
That would be a nice feature. Any updates on this?

PR #610, opened 2024-10-28
Handle phone numbers that start with "%2B"

crummy:
# Problem

I've generated a Java SDK for the signal-cli-rest-api with liblab. They parse their URL arguments with `URLEncoder.encode(value, StandardCharsets.UTF_8.toString());`, which turns `+1234` into `%2B1234`. Thus, when I use the client to query the API, I get a 400 error because `/v1/groups/%2B1234` is malformed.

# Solution

Escape the number when parsing it.

# Comments

I have never written Go code before, so feel free to point out any problems here!

Googling, it seems like `+` is actually allowed in part of the URL - it must be escaped if following a `?`, and can be escaped before. So it seems reasonable to handle both ways.

### DIFF
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -144,9 +144,9 @@ type CreateGroupResponse struct {
 }
 
 type UpdateProfileRequest struct {
-	Name         string `json:"name"`
-	Base64Avatar string `json:"base64_avatar"`
-	About 			 *string `json:"about"`
+	Name         string  `json:"name"`
+	Base64Avatar string  `json:"base64_avatar"`
+	About        *string `json:"about"`
 }
 
 type TrustIdentityRequest struct {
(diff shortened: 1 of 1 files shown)


src/api/api.go
```diff
@@ -232,7 +233,10 @@ func (a *Api) About(c *gin.Context) {
 // @Param data body RegisterNumberRequest false "Additional Settings"
 // @Router /v1/register/{number} [post]
 func (a *Api) RegisterNumber(c *gin.Context) {
-	number := c.Param("number")
+	number, err := url.PathUnescape(c.Param("number"))
+	if err != nil {
+		c.JSON(400, Error{Msg: "Couldn't process request - malformed number"})
+	}
```
bbernhard, line 239: Thanks for the PR!

In general it looks good, the only thing that is missing is a `return` in case a malformed number is provided.

crummy: Thanks, fixed.

bbernhard:
Thanks a lot!

PR #612, opened 2024-11-09
Add notify-self-flag support for json-rpc mode

ahatius:
Relates to #573, unfortunately I forgot to add the flag for the json-rpc mode which this PR will do. Should work the same, if the flag is not set, it should enable the self-notification by default so it is consistent with the existing behavior.


bbernhard:
looks good, thanks!

PR #633, opened 2024-12-24
added support for Grafana and Graylog alerting

hvdort:
Added support for 2 additional api calls:

Grafana (and other AlertManager based alerting tools):

/v2/sendalertmanager

No special things to do. Just create a Contactpoint of type Webhook (POST) with url pointed to http://<servername>:8485/v2/sendalertmanager

Graylog:

/v2/sendgraylognotification

mandatory fields in Event Definition:
- fromnumber
- message
- recipients

### DIFF
--- a/src/api/graylogapi.go
+++ b/src/api/graylogapi.go
@@ -1,0 +1,211 @@
+package api
+
+import (
+	"encoding/json"
+	"errors"
+	"strconv"
+	"strings"
+	"fmt"
+
+	_ "runtime/debug"
+	_ "github.com/yassinebenaid/godump"
+
+	"github.com/gin-gonic/gin"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/bbernhard/signal-cli-rest-api/client"
+)
+
+type AlertManagerNotification struct {
+	Receiver          string      `json:"receiver"`
+	Status            string      `json:"status"`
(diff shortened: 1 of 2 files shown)


bbernhard:
Thanks a lot for the contribution!

I really thought long about your PR and finally decided to not integrate it into the codebase. It's not about the quality of the PR, but more that those endpoints are only useful for a small number of people. And since there are a lot of different monitoring/alerting systems out there, which probably all require different API structures, we would end up with a lot of different API endpoints.

Personally I am a bit afraid of the maintenance effort and the clutter it will create in the swagger documentation. I think the core of this project should be kept as simple as possible. Nevertheless, I understand, that people want to use the API for all different tasks, which can't be accomplished due to the fixed and inflexible structure of the API. 

Therefore I decided to implement a Plugin mechanism. This Plugin mechanism allows you to dynamically register custom API endpoints with arbitrary payloads. Plugins consist of a definition file and a lua script file. [Have a look here](https://github.com/bbernhard/signal-cli-rest-api/tree/master/plugins) for details. 

The plugin mechanism will be part of the next release, but I can provide you a docker image for testing if you are interested. The idea is that people then can write their own plugins to add additional API endpoints. Any plugin contributions are greatly appreciated - so please feel free to create a PR :) 

If you have any questions or need help, please let me know.

Issue #651, opened 2025-02-05
Unable to link device via QR code


hillct:
### The problem

As of January 24, I am no longer able to perform QR code based device linking, from the android Signal client, which now reports that the QR code created by signal-cli-rest-api does not represent a valid signal linking URI. I'm guessing the issue arose on the 24th, because that was the release date of the signal android client v7.31.1 however, I can't say for sure, as my test server remained linked until the 28th (when I unlinked). I have since not been able to re-link. An image of the error as presented in the android signal client is attached. No errorts are logged on the signal-cli-rest-api side, beyond the informational QR code http request. I would expect this related to some slight restructuring of linking URLs on the signal side which would likely have to be addressed upstream, within https://github.com/AsamK/signal-cli but figured it would be worth noting here.

![Image](https://github.com/user-attachments/assets/96a7041e-2bf3-4ae4-82a1-3abc5c8a78ea)

### Are you using the latest released version?

- [x] Yes

### Have you read the troubleshooting page?

- [x] Yes

### What type of installation are you running?

signal-cli-rest-api Docker Container

### In which mode are you using the docker container?

Normal Mode

### What's the architecture of your host system?

arm64

qianmianyao:
I'm having the same problem.

bbernhard:
Does it make a difference if you try another mode (e.g `json-rpc`)? 

Just tried with Version 7.32.2 on my Android phone and it works there.

hillct:
@bbernhard Thanks for your quick followup. I've performed the tests you suggest. I was already using v0.178-dev [v0.178-dev](https://hub.docker.com/layers/bbernhard/signal-cli-rest-api/0.178-dev/images/sha256-7a9cf2e171c3e7e8e2bdd8a290b45a8b84f44b3442db850b1e92213780d013a0) so updated my android signal client to 7.32.2 with the server in normal mode, with the same results. I then switched to json-rpc mode which yielded the following error at `http://localhost:8080/v1/qrcodelink?device_name=hobart-signal-bot`

And the following docker console log:
```
Attaching to signal-cli-rest-api-admin
signal-cli-rest-api-admin  | + set -e
signal-cli-rest-api-admin  | + [ -z /home/.local/share/signal-cli ]
signal-cli-rest-api-admin  | + usermod -u 1000 signal-api
signal-cli-rest-api-admin  | + groupmod -o -g 1000 signal-api
signal-cli-rest-api-admin  | usermod: no changes
[...]

PR #652, opened 2025-02-08
Corrected for JVM crash by using more robust LTS JRE base image, sabi…

hillct:
Corrected for JVM crash by using more robust LTS JRE base image, saving 200MB in final image size as a bonus.
This work was done in an attempt to address https://github.com/bbernhard/signal-cli-rest-api/issues/651 which it does not do completely, but the enhancements to the Dockerfile are generally beneficial in that they increase build predictability by shifting to the LTS OpenJRE image.

This has not been tested across all supported platform as I don't have any arm7 hardware with which to test. It should be noted that arm7 continues to utilize the old Jammy image, so it should be fine anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,109 +24,110 @@ COPY ext/patches/signal-cli-native-arch.patch /tmp/signal-cli-native-arch.patch
 RUN arch="$(uname -m)"; \
         case "$arch" in \
             aarch64) cp /tmp/libsignal-client-libraries/arm64/libsignal_jni.so /tmp/libsignal_jni.so ;; \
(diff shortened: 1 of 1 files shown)


bbernhard:
I really appreciate the effort you put into this! However, I am a bit worried that having two different base images will increase the maintenance burden significantly. I had something like this in the early stages of this project and it was a bit of a pain to maintain and resulted in some subtle bugs. :/ 

I asked the eclipse-temurin guys a while ago whether they can re-enable the armv7 builds and I think there were also some active discussions about that, but I am not sure if there was actually some progress there. Need to look up that issue in their issue tracker.

That being said, I would definitely be more comfortable, if there would just be one base image for all the supported architectures. But I'll think about it a bit more.

reneleonhardt:
Wow, amazing work @hillct 🚀

PR #663, opened 2025-02-25
Path parameter in /v1/search is required

crummy:
# Problem

editor.swagger.io reports this error with the swagger.yml:
```
Structural error at paths./v1/search/{number}.get.parameters.0
should have required property 'required'
missingProperty: required
```

# Solution

Set the `required` parameter to `true` for the `/v1/search` endpoint.

# Comments

I tried to regenerate all the docs but get a bunch of other unnecessary changes so I'll leave that up to you @bbernhard - sorry!


bbernhard:
looks good, thanks!

Issue #706, opened 2025-06-10
Updating the Home-Assistant doc: Solving CAPTCHA issues when sending messages to new recipient via REST-API on Home Assistant


wassong:
Hello everyone,

I ran into some issues while trying to install and use Signal on Home Assistant via the REST API. I’d like to share what went wrong and how I was able to fix it.

## Problem

When trying to send a message to a new Signal account, I received the following error:

`Failed to send (some) messages: <receiving number>: CAPTCHA proof required for sending to "<receiving number>", available options "PUSH_CHALLENGE, CAPTCHA" with challenge token "<challenge>", or wait "86400" seconds. To get the captcha token, go to https://signalcaptchas.org/challenge/generate.html After solving the captcha, right-click on the "Open Signal" link and copy the link. Use the following command to submit the captcha token: signal-cli submitRateLimitChallenge --challenge CHALLENGE_TOKEN --captcha CAPTCHA_TOKEN`

## My Steps to Resolve It

1. I registered a new landline number for use with Signal.
2. When sending a message to a number that hadn’t interacted with my instance before, I got the CAPTCHA error above.
3. I visited [https://signalcaptchas.org/challenge/generate.html](https://signalcaptchas.org/challenge/generate.html), solved the CAPTCHA, and copied the URL behind the **"Open Signal"** button.
4. I used that link in the following REST call:

```bash
curl -X POST \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
        "captcha": "signalcaptcha://<captcha>", 
        "challenge_token": "<challenge>"
      }' \
  'http://<ip>:<port>/v1/accounts/<sending number>/rate-limit-challenge'
```

## Important Note
You will encounter this CAPTCHA challenge for each new number you send a message to for the first time. That means you have to repeat steps 3 and 4 for every new recipient.

Xopher00:
Can I ask how you got this working? I have encountered the rate limit challenge. I am going thru every step, copying the challenge token and the captcha. I have tried this several times now and I keep getting a 400 error but no details. It is incredibly frustrating

PR #727, opened 2025-07-22
Update captcha instructions in HOMEASSISTANT.md

jakommo:
I had trouble with the current instructions to find the correct link in dev tools. After seeing the instructions is https://github.com/bbernhard/signal-cli-rest-api/issues/706 I could easily get it to work. 

Updating the instructions here per issue above, as this seems much more accessible / easy than trying to find this in dev tools. 

cc @wassong since you shared those in https://github.com/bbernhard/signal-cli-rest-api/issues/706 (thank you!)


bbernhard:
Looks good, many thanks!

PR #735, opened 2025-08-11
Allow the startup chown to be optional

WeldFire:
Currently, the `entrypoint.sh` script unconditionally runs `chown -R` on the `${SIGNAL_CLI_CONFIG_DIR}` during every startup. While this ensures permissions are correct, it can cause extremely long startup times on systems using network-attached storage (NFS) or other non-local volume mounts.

For users with large data directories on these systems, this can increase the container startup time from a few seconds to several minutes, even when permissions are already correct.

#### What is the solution?

This pull request introduces a new environment variable, `SIGNAL_CLI_CHOWN_ON_STARTUP`, to make this operation optional.

- The `entrypoint.sh` script is modified to only run the `chown -R` command if `SIGNAL_CLI_CHOWN_ON_STARTUP` is not set to `'false'`.
- The `Dockerfile` sets a default value of `true` for this variable to maintain existing behavior and ensure backward compatibility for all users.
- The `README.md` is updated to document the new variable and its usage.


bbernhard:
Looks good, thanks!

PR #741, opened 2025-08-29
docs: add contributing.md and additional instructions on building the docs

Gara-Dorta:
I found it quite hard to get started with setting up a dev environment so I wrote what I found in the hope that it'll be helpful to others. Probably not the best way to go about it. So please let me know and I'll update accordingly.

### DIFF
--- a/src/docs/README.md
+++ b/src/docs/README.md
@@ -18,12 +18,12 @@ Then run the app in `/src`
 go run main.go
 ```
 
-Or with the docker compose in the root of the repository
+Or with docker compose in the root of the repository
 ```bash
 docker compose up
 ```
 
-And navigate to the following address to view the docs
+Then navigate to the following address to view the docs
 
 http://127.0.0.1:8080/swagger/index.html
 


bbernhard (commented):
Thanks a lot! Documentation is always something that is highly appreciated!

What I usually do (not sure though if it is more intuitive than your approach), is the following: 

I change this [docker-compose.yml](https://github.com/bbernhard/signal-cli-rest-api/blob/master/docker-compose.yml) file and remove the `image: bbernhard/signal-cli-rest-api:latest` line and replace it with `build: "."`

Then my workflow looks like this:

* change something in the sourcecode
* run `docker compose build` to build the docker image
* run `docker compose up` to start the docker container
* test the changes
* run `docker compose down`
* do again some changes
* build + run again

That way one doesn't need to care about tagging the docker image nor setting the `BUILD_VERSION` is necessary.

But again, that's just how I usually do it. For me that feels a bit more natural :)

Gara-Dorta:
Your approach is a lot simpler.  When I tried it, I got errors building the image because it couldn't get internet access. That's why I had to add the `--network=host`. I'll update the instructions to follow your steps.

bbernhard:
Many thanks!

PR #770, opened 2025-12-05
Improve captcha documentation

jtrumpio:
Hi, I wanted to improve the documentation to show an example captcha string.  I ended up rate limiting myself by accident before I figured out what the captcha string needed to look like.

Thank you!

Jason


bbernhard:
Thanks!

PR #772, opened 2026-01-08
New/changes endpoints for linking devices

Kamaradas:
This PR introduces improvements to the device linking workflow, including updates to existing endpoints and the addition of new ones to support more flexible device management.

Changed Endpoints

	•	/v1/devices/{number} (ListDevices)
The endpoint now returns the device ID, which is required by newly introduced endpoints.

New Endpoints

	•	/v1/qrcodelink/raw (GetQrCodeLinkUri)
Returns a device link URI that can be used directly by the device linking endpoint. Previously, the URI link was only possible via a QR code.

	•	/v1/devices/{number}/{deviceId} (RemoveDevice)
Removes a linked device from the master device using the device ID returned by ListDevices.

	•	/v1/devices/{number}/local-data (DeleteLocalAccountData)
Deletes the local database from the device. This is typically used after a device has been removed from the master device.

Testing

These changes were tested on ARM and x86_64 architectures, and in both normal and JSON-RPC modes.

### DIFF
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -220,6 +220,14 @@ type RemoteDeleteRequest struct {
 	Timestamp int64  `json:"timestamp"`
 }
 
+type DeleteLocalAccountDataRequest struct {
+    IgnoreRegistered bool `json:"ignore_registered" example:"false"`
+}
+
+type DeviceLinkUriResponse struct {
+    DeviceLinkUri string `json:"device_link_uri"`
+}
+
 type Api struct {
 	signalClient *client.SignalClient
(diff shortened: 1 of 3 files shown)


bbernhard:
Many thanks - very much appreciated!

Kamaradas:
Do you know when you will release the new version with these changes?

bbernhard:
> Do you know when you will release the new version with these changes?

Hopefully soon. :)  If you are interested in an image with the changes, you could e.g use this image: `bbernhard/signal-cli-rest-api:1768342695-ci`

PR #787, opened 2026-02-13
ListContact Improvements

dangnabit:
`client.ListContact(s)` currently only returns data about an account's contact list. This change extends that to use the `--all-recipients` option from the `signal-cli`, returning known recipient (co-members of large groups) information instead of only contacts. 

**Details:**
- Adds `all_recipients` boolean query param
- Combines `ListContact` and `ListContacts` in the client, allowing the CLI to perform the search instead of requesting all contacts and finding the matching contact. 
- Updated swagger with `swag init`, which also picked up a previous change to `GroupPermissions`

### DIFF
--- a/src/client/client.go
+++ b/src/client/client.go
@@ -214,31 +214,6 @@ type ContactProfile struct {
 	LastUpdatedTimestamp int64  `json:"last_updated_timestamp"`
 }
 
-type SignalCliProfileResponse struct {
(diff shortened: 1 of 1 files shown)


skleinjung:
I'm definitely looking for this functionality. Thanks for the patch!

bbernhard:
Many thanks!

Issue #796, opened 2026-02-26
Incomplete JSON serialization for incoming Reactions (message: null, missing reaction object)


sandman12345678:
### The problem

When receiving a reaction (emoji) from an external contact via the /v1/receive endpoint in MODE=normal, the API returns a dataMessage envelope where message is null, but the expected reaction object (containing emoji, targetSentTimestamp, etc.) is entirely missing from the JSON output.

To Reproduce

    Run the API in MODE=normal.

    Send a text message from an external Signal account to the API number.

    Call GET /v1/receive/+<number>.

    While the poll is active, react to the message from the external account with an emoji (e.g., 👍).

    Observe the JSON output.

Actual Behavior
The API returns a valid JSON structure for the event, but the dataMessage is empty except for the timestamp and null-message:
JSON

{
  "envelope": {
    "source": "+49123456789",
    "sourceNumber": "+49123456789",
    "timestamp": 1772140664347,
    "dataMessage": {
      "timestamp": 1772140664347,
      "message": null,
      "expiresInSeconds": 0,
      "isExpirationUpdate": false,
      "viewOnce": false
    }
  }
}

Expected Behavior
The dataMessage should contain a reaction object, similar to how the /v1/reactions POST endpoint structures its input:
JSON

"dataMessage": {
  "timestamp": 1772140664347,
  "message": null,
  "reaction": {
    "emoji": "👍",
    "targetAuthor": "+49987654321",
    "targetSentTimestamp": 1772140661000,
    "isRemove": false
  }
}

Context

    Mode: normal (Native)

    Version: [Deine Version, z.B. 0.7.x]

    Note: Normal text messages are received correctly. Only reactions seem to lose their payload during JSON serialization.

### Are you using the latest released version?

- [x] Yes

### Have you read the troubleshooting page?

- [x] Yes

### What type of installation are you running?

signal-cli-rest-api Docker Container

### In which mode are you using the docker container?

Normal Mode

### What's the architecture of your host system?

x86-64

bbernhard:
I think that is an upstream issue (could be related to this one: https://github.com/AsamK/signal-cli/issues/1935)

strichter:
This is related to: https://github.com/AsamK/signal-cli/issues/1940

strichter:
Best discussion is here: https://github.com/adstastic/nanoclaw/pull/15

strichter:
Issue fixed here: https://github.com/AsamK/signal-cli/pull/1944

[...]

PR #799, opened 2026-02-28
Fix Signal Desktop v8.0.0 binary ACI compatibility

strichter:
## Summary

Signal Desktop v8.0.0 switched from string ACI fields (`mentionAci`, `authorAci`, `targetAuthorAci`) to binary ACI encoding (`mentionAciBinary`, `authorAciBinary`, `targetAuthorAciBinary`) in protobuf messages. This causes:

- **Reactions**: target author resolves to null → empty reaction messages
- **Mentions**: ServiceId is null → mentions silently dropped
- **Quotes**: author is null/invalid → entire quote dropped
- **Sync messages**: missing destination validation failures

### Fix

Two-part patch applied to signal-cli v0.13.24 during the x86_64 source build:

1. **Bump `signal-service-java`** from `unofficial_137` → `unofficial_138` — adds dual-format ACI parsing (`ServiceId.parseOrNull(stringField, binaryField)`) so both old string and new binary formats work.

2. **Defensive null guards in `MessageEnvelope.java`** — for edge cases where ServiceId resolution still returns null/`ACI.UNKNOWN`:
   - Reactions: filter null `getTargetAuthor()` 
   - Mentions: filter null `getServiceId()`
   - Quotes: fallback to `UNKNOWN_UUID` instead of dropping the quote
   - StoryContext/PollVote: null guards on author fields

### How it works

- On x86_64: the source build applies `ext/patches/fix-binary-aci.patch` to the v0.13.24 checkout, then `installDist` replaces the release tarball directory. Both the JVM and GraalVM native paths get the fix.
- On arm64/armv7: continues using the unpatched release tarball (no regression). These architectures will get the fix when signal-cli cuts a new release with `unofficial_138`.

### Verified

Tested end-to-end with Signal Desktop v8.0.0 sending to a linked signal-cli instance:
- ✅ Reactions from Desktop: `targetAuthorUuid` populated
- ✅ Mentions from Desktop: `mentions` array with UUIDs present  
- ✅ Quotes from Desktop: `quote.authorUuid` populated

### Related

- signal-cli PR: https://github.com/AsamK/signal-cli/pull/1944
- signal-service-java `unofficial_138`: dual-format ACI parsing

## Test plan

- [x] Build Docker image on x86_64 (`podman build .`)
- [x] Verify signal-cli processes reactions from Signal Desktop v8.0.0
- [x] Verify signal-cli processes mentions from Signal Desktop v8.0.0
[...]

PR #801, opened 2026-02-28
Fix Signal Desktop v8.0.0 binary ACI by building signal-cli from master

strichter:
## Summary

Signal Desktop v8.0.0 switched from string ACI fields to binary ACI encoding in protobuf messages, breaking reactions, mentions, and quotes in signal-cli v0.13.24. The signal-cli maintainer [confirmed](https://github.com/AsamK/signal-cli/pull/1944) that the master branch already includes the fix (`signal-service-java:2.15.3_unofficial_138` with dual-format ACI parsing).

This PR builds signal-cli from master on x86_64 instead of checking out the v0.13.24 tag, which resolves all three issues without needing a patch file:

- **Reactions** from Desktop: `targetAuthorUuid` populated
- **Mentions** from Desktop: `mentions` array with resolved UUIDs
- **Quotes** from Desktop: `quote.authorUuid` populated

### Changes

- Build from signal-cli master (default branch) instead of `v0.13.24` tag on x86_64
- Use `installDist` output as the main signal-cli installation (replaces the release tarball on x86_64)
- Detect `libsignal-client` jar version dynamically (master uses 0.87.1 vs 0.87.0 in v0.13.24)
- Update `signal-cli-native.patch` to match master's `build.gradle.kts` structure
- Move BaseConfig/libsignal workarounds behind architecture guard (arm64/armv7 only — they still use the v0.13.24 release tarball)

### How it works

- **x86_64**: Source build clones signal-cli master, builds with `./gradlew installDist`, replaces the release tarball directory, injects the custom `libsignal_jni.so`, then compiles the GraalVM native image.
- **arm64/armv7**: No change — continues using the v0.13.24 release tarball with existing workarounds. These architectures will get the fix when signal-cli cuts a new release.

Supersedes #799 (which patched v0.13.24 with the same fixes that are now on master).

## Test plan

- [x] Verified signal-cli master handles binary ACI correctly (tested reactions, mentions, quotes from Signal Desktop v8.0.0 to a linked signal-cli instance)
- [ ] Build Docker image on x86_64 (`docker build .`)
- [ ] Verify arm64/armv7 builds are unaffected (no changes to those code paths)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -167,7 +167,7 @@ RUN cd /tmp/signal-cli-rest-api-src && go build -buildmode=plugin -o signal-cli-
 # is fixed we use the standard ubuntu image
 #FROM eclipse-temurin:21-jre-jammy
 
-FROM ubuntu:jammy
+FROM ubuntu:noble
 
 ENV GIN_MODE=release
 
@@ -181,7 +181,7 @@ ENV SIGNAL_CLI_REST_API_PLUGIN_SHARED_OBJ_DIR=/usr/bin/
 
 RUN dpkg-reconfigure debconf --frontend=noninteractive \
 	&& apt-get update \
-	&& apt-get install -y --no-install-recommends util-linux supervisor netcat openjdk-21-jre curl locales \
+	&& apt-get install -y --no-install-recommends util-linux supervisor netcat-openbsd openjdk-25-jre curl locales \
 	&& rm -rf /var/lib/apt/lists/* 
 
 COPY --from=buildcontainer /tmp/signal-cli-rest-api-src/signal-cli-rest-api /usr/bin/signal-cli-rest-api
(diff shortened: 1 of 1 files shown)


strichter:
Fixes #796 — builds signal-cli from master which includes the binary ACI fix, resolving incomplete JSON serialization for reactions, mentions, and quotes from Signal Desktop v8.0.0. Supersedes #799.

bbernhard:
Thanks for the PR, but I do not want to include any patches for signal-cli and/or build signal-cli directly from the master branch. Building from master could be quite dangerous, since it might contain some partial commits.

TWhidden:
Since our bot depends on the websocket events from this repos API, the missing mentions field is currently causing failures on our end. @bbernhard do you have a rough idea when this repository might be updated to include the fixed signal-cli version (or otherwise get this change merged)? A timeline, even very approximate would help us plan.

bbernhard:
> Since our bot depends on the websocket events from this repos API, the missing mentions field is currently causing failures on our end. @bbernhard do you have a rough idea when this repository might be updated to include the fixed signal-cli version (or otherwise get this change merged)? A timeline, even very approximate would help us plan.

Which architecture are you using? The code in the repository should already build for x86-64 (and probably for armv7 too). So, if it's one of those architectures, you should be good to go :)

PR #808, opened 2026-03-06
Enable verbose logging in `signal-cli` if log level is debug

EvanHahn:
*Feel free to reject if this isn't desired.*

If the log level is `debug` (either with the `LOG_LEVEL` environment variable or configuration option), enable verbose logging in `signal-cli`.

I thought it'd be useful to enable verbose logging in `signal-cli`. This does that automatically.

### DIFF
--- a/src/client/cli.go
+++ b/src/client/cli.go
@@ -89,6 +89,10 @@ func (s *CliClient) Execute(wait bool, args []string, stdin string) (string, err
 		args = append([]string{"--trust-new-identities", trustModeStr}, args...)
 	}
 
+	if log.GetLevel() >= log.DebugLevel {
+		args = append([]string{"--verbose"}, args...)
+	}
+
 	fullCmd := ""
 	if stdin != "" {
 		fullCmd += "echo '" + stdin + "' | "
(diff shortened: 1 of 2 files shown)


bbernhard:
Thanks for the PR!

The problem I see here is, that the additional output that is returned breaks actual REST API endpoint. So, while additional debug info is logged, the actual REST API endpoint might not return any data anymore (as the data can't be parsed anymore).

EvanHahn:
Good point. Could this be added to JSON RPC mode (maybe with a separate environment variable)? Again, feel free reject if this isn't desired.

bbernhard:
I think in `json-rpc` mode the logs are written to `stdout`, so we don't see them. But it looks like that `signal-cli` offers the possibility to log to a file:

> --log-file LOG_FILE
> Write log output to the given file. If --verbose is also given, the detailed logs will only be written to the log file.

So, I guess what would work is:

* Introduce a new `verbose` log level
* if the `verbose` log level is set, the `signal-cli` binary is also called with the `--log-file` parameter

Since we are logging into a file then, I _think_ that should then also for the `normal` and `native` mode without problems.

Would that be fine for your use case?

EvanHahn:
I'll check on that and get back to you.

Issue #723, opened 2025-07-07
Disable automatic file downloads (such as attachments, avatars, or stickers)


Optega:
I need to use the API only for text messages. Automatic file downloads (such as attachments, avatars, or stickers) create unnecessary load and consume bandwidth, as all received media files are saved by default in `/home/.local/share/signal-cli`.

Since I can't restrict this on the user side, I’d like to control it on the server side instead.
Is there a way to disable automatic downloading, or could such a feature be added in the future — for example, in the [Advanced Settings](https://github.com/bbernhard/signal-cli-rest-api?tab=readme-ov-file#advanced-settings)?

bbernhard:
For which mode (json-rpc, normal or native) would you like to have that?

Optega:
> For which mode (json-rpc, normal or native) would you like to have that?

I am using json-rpc mode

bbernhard:
Should be implemented and will be released with the next version. The behavior can be configured by setting the `JSON_RPC_IGNORE_ATTACHMENTS` (and `JSON_RPC_IGNORE_STORIES`)  env variables in the `docker-compose.yml` to `true`. If you want to give it a try, use this test version: `bbernhard/signal-cli-rest-api:0.193-dev`

olehrb1397:
Hi @bbernhard 
Great job on the project, it has been really helpful for us so far.

Is there any kind of an expected release date for those features?

We are hosting the Signal API in json-rpc mode in GCP just to send messages with attachments to a particular group, and there we are not actually interested in incoming messages, just need the "receive" functionality to ensure, when a new user is added to the target group(s) they start seeing new messages as soon as possibly.

Currently we are experiencing an issue, where we ran out of ROM, due to the amount of downloaded attachments, therefore this fix is quite crucial for us.

bbernhard:
> Hi [@bbernhard](https://github.com/bbernhard) Great job on the project, it has been really helpful for us so far.
> 
> Is there any kind of an expected release date for those features?
> 
> We are hosting the Signal API in json-rpc mode in GCP just to send messages with attachments to a particular group, and there we are not actually interested in incoming messages, just need the "receive" functionality to ensure, when a new user is added to the target group(s) they start seeing new messages as soon as possibly.
> 
> Currently we are experiencing an issue, where we ran out of ROM, due to the amount of downloaded attachments, therefore this fix is quite crucial for us.

Thanks for the kind words!

New release is out.

Optega:
[...]

PR #809, opened 2026-03-07
Add --ignore-avatars and --ignore-stickers support

### LINKED ISSUE #776: Can we extend ignoring files in JSON-RPC to include stickers and avatars?
In #723 great work was done on attachments and stories.  I would like to keep a long running instance and would prefer to not pollute a backup with stickers and avatars I don't need.  It isn't clear to me if the content of data/ is ephemeral or not.  Basically I'd like the leanest text only set up possible.

If this is acceptable and I get a bit of advice I can look at trying to put together a PR for this if it is within my skill set.

bexelbie:
Expose the signal-cli v0.14.0 `--ignore-avatars` and `--ignore-stickers` flags, following the existing `--ignore-attachments`/`--ignore-stories` pattern.

**JSON-RPC mode** (env vars for daemon startup):
- `JSON_RPC_IGNORE_AVATARS=true`
- `JSON_RPC_IGNORE_STICKERS=true`

**Normal/Native mode** (query parameters on `/v1/receive`):
- `ignore_avatars=true`
- `ignore_stickers=true`

**Auto-receive scheduler** (env vars):
- `AUTO_RECEIVE_SCHEDULE_IGNORE_AVATARS=true`
- `AUTO_RECEIVE_SCHEDULE_IGNORE_STICKERS=true`

Upstream commit: AsamK/signal-cli@fefca7d837a12975de7147f49712eb16fd6407c4

Fixes #776, relates to #723

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,24 +1,11 @@
 services:
   signal-cli-rest-api:
-    #image: bbernhard/signal-cli-rest-api:latest-dev
-    build: "."
+    image: bbernhard/signal-cli-rest-api:latest
     environment:
-      - MODE=normal #supported modes: json-rpc, json-rpc-native, native, normal
-      - ENABLE_PLUGINS=true
-      - DEFAULT_SIGNAL_TEXT_MODE=styled
-      - SWAGGER_IP=127.0.0.1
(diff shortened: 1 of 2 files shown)


bbernhard:
Many thanks!

PR #833, opened 2026-04-08
Mark name and members as optional fields in CreateGroupRequest

Gara-Dorta:
Following the discussions in https://github.com/bbernhard/signal-cli-rest-api/pull/828#discussion_r3047189125 this PR marks the `Name` and `Members` arguments of the `CreateGroupRequest` as optional.

For the native mode, luckily, signal-cli would just accept being called with `signal-cli updateGroup -n -m` and create an empty group. However, instead of relying on that behaviour to continue to work in the future this PR avoids calling `signal-cli` with empty parameters.

### DIFF
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -39,8 +39,8 @@ type UpdateContactRequest struct {
 }
 
 type CreateGroupRequest struct {
-	Name           string              `json:"name"`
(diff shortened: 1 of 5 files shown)


bbernhard:
Thanks!

Issue #255, opened 2022-07-13
Couldn't send message to golang channel, as there's no receiver


asad-awadia:
### The problem

getting this when sending a message via v2/send using json-rpc mode?

```
Couldn't send message to golang channel, as there's no receiver
```

@bbernhard 

### Are you using the latest released version?

- [X] Yes

### Have you read the troubleshooting page?

- [X] Yes

### What type of installation are you running?

signal-cli-rest-api Docker Container

### In which mode are you using the docker container?

JSON-RPC Mode

### What's the architecture of your host system?

x86-64

asad-awadia:
getting 400 on sending a message?

 no logs why

bbernhard:
Can you please connect to the docker container with:

`docker exec -it <container id> /bin/bash`

and have a look inside the `/var/log/signal-cli-json-rpc` folder. Do you see any errors there?

Does the `native`/`normal` mode work?

asad-awadia:
@bbernhard 
```
WARN  ManagerImpl - No profile name set. When sending a message it's recommended to set a profile name wit the updateProfile command. This may become mandatory in the future.
WARN  ManagerImpl - No profile name set. When sending a message it's recommended to set a profile name wit the updateProfile command. This may become mandatory in the future.
WARN  ManagerImpl - No profile name set. When sending a message it's recommended to set a profile name wit the updateProfile command. This may become mandatory in the future.
WARN  ManagerImpl - No profile name set. When sending a message it's recommended to set a profile name wit the updateProfile command. This may become mandatory in the future.
WARN  ReceiveHelper - Connection closed unexpectedly, reconnecting in 100 ms
```

just lots of these


craziest part is that for 1 number messages don't go at all

and for 1 type of number only some messages go

but for my number all messages go

asad-awadia:
starting working on its own - closing issue

PR #835, opened 2026-04-08
feat(jsonrpc2): add manual receive mode to prevent message loss

### LINKED ISSUE #834: JSON-RPC mode: messages dropped when no websocket subscriber is attached
## Summary

In `MODE=json-rpc`, the `/v1/receive/{number}` websocket endpoint silently drops every inbound message that arrives while no subscriber is currently reading from its channel. signal-cli has already pulled the message from the Signal servers at that point, so it is never redelivered. This is the same root cause behind #255 and #185 (and the same `Couldn't send message to golang channel, as there's no receiver` debug log line).

The most painful manifestation is during a deploy/restart of the websocket client: the brief gap between the old client disconnecting and a new client subscribing eats every message that arrives in that window. With a 60-second client outage, **100% of messages sent during that window are lost**.

## Root cause (current behaviour, json-rpc mode)

`signal-cli` is launched in the default `--receive-mode=on-start`, so the daemon eagerly pulls inbound messages from the Signal servers and pushes them as JSON-RPC notifications, regardless of whether any websocket subscriber is connected.

When a notification arrives at `signal-cli-rest-api`, `ReceiveData` in `src/client/jsonrpc2.go` does a non-blocking broadcast to all registered subscriber channels:

```go
for _, c := range r.receivedMessagesChannels {
    select {
    case c <- resp1:
        log.Debug("Message sent to golang channel")
    default:
        log.Debug("Couldn't send message to golang channel, as there's no receiver")
    }
}
```

If `receivedMessagesChannels` is empty (the websocket client just disconnected) or every channel's buffer is full, the `default` branch fires and the message is silently discarded with only a `Debug`-level log line. signal-cli has already acked the message to Signal at that point, so the Signal servers will not redeliver.

## Reproduction

This was reproduced end-to-end with a real Signal account against a real `signal-cli-rest-api` container. The harness:

1. Stops the websocket client.
2. Sends N messages from a separate Signal account to the bot's number.
3. Waits 60 seconds.
4. Starts the websocket client.
5. Counts how many of the N messages were delivered.

**Result against unpatched master**: 0 of 10 messages delivered (100% loss).

## Proposed fix

signal-cli already supports a `--receive-mode=manual` mode that is exactly designed for this scenario: signal-cli does NOT auto-pull messages, and clients explicitly call `subscribeReceive` to start receiving. While no subscription is active, messages stay buffered server-side under Signal's normal retention rules.

The fix in `signal-cli-rest-api`:

1. Add a `JSON_RPC_RECEIVE_MODE` environment variable. Default unset = current behaviour (auto). `JSON_RPC_RECEIVE_MODE=manual` opts in.
2. When set to `manual`, `jsonrpc2-helper.go` adds `--receive-mode=manual` to the signal-cli daemon command line.
3. In `JsonRpc2Client`, add per-account subscription tracking with refcounting. When the first websocket subscriber for an account attaches, call `subscribeReceive(account)`. When the last subscriber for that account detaches, call `unsubscribeReceive(account, subscriptionId)`.
4. In `ReceiveData`, unwrap the manual-mode `{subscription, result}` notification format so downstream consumers (the websocket handler) see the same envelope shape in both modes.
5. Thread the `account` parameter through `SignalClient.GetReceiveChannel(number)` and `JsonRpc2Client.GetReceiveChannel(account)` so the per-account subscription can be acquired/released at attach/detach time.

The change is fully backward compatible: when `JSON_RPC_RECEIVE_MODE` is unset or set to `on-start`, the new code path is bypassed and behaviour is unchanged.

## Validation

I built the patched image locally, swapped it in for `signal-cli-rest-api` in a real nora bot deployment, and re-ran the same harness:

| Configuration | Phase 4 (60s outage, 10 msgs sent during gap) |
|---|---|
| Master, default `on-start` | **10/10 lost (100% loss)** |
| Patched, `JSON_RPC_RECEIVE_MODE=manual` | **0/10 lost (0% loss)** |

The trace in the signal-api log clearly shows the new lifecycle:

```
20:22:25 Subscribed to receive notifications for account +1XXXXXXXXXX (subscription=0)
20:22:47 Unsubscribed from receive notifications for account +1XXXXXXXXXX (subscription=0)
20:28:32 Subscribed to receive notifications for account +1XXXXXXXXXX (subscription=1)
```

Between 20:22:47 (unsubscribe) and 20:28:32 (resubscribe), 10 messages were sent into the group. After resubscribe, all 10 were delivered to the websocket client over the next ~10 seconds.

## PR

I will follow up with a PR shortly. Branch is at: https://github.com/strichter/signal-cli-rest-api/tree/feat/jsonrpc2-manual-receive-mode-upstream

Happy to discuss design / split into smaller commits / address concerns about the refcounting model before opening the PR.

strichter:
Fixes #834. Same root cause as the long-standing #255 (\"Couldn't send message to golang channel, as there's no receiver\").

## Summary

In \`MODE=json-rpc\` with default \`--receive-mode=on-start\`, signal-cli auto-pulls inbound messages and pushes them as JSON-RPC notifications regardless of whether any websocket subscriber is currently attached. If no subscriber is reading the channel at the moment a notification arrives, the receiver loop in \`src/client/jsonrpc2.go\` drops the message via a non-blocking channel send and only logs a debug line. signal-cli has already acked the message to Signal at that point, so it is never redelivered.

This PR adds opt-in support for signal-cli's \`--receive-mode=manual\`, where the daemon does NOT auto-receive and clients must explicitly call \`subscribeReceive\` to start receiving. While no subscription is active, messages stay queued server-side under Signal's normal retention rules.

See #834 for the full reproduction, root cause analysis, and validation results.

## Changes

- **\`src/scripts/jsonrpc2-helper.go\`**: thread the new \`JSON_RPC_RECEIVE_MODE\` env var into the signal-cli daemon command line. Defaults unset / \`on-start\` to current behaviour. Setting it to \`manual\` adds \`--receive-mode=manual\` to the launch.
- **\`src/client/jsonrpc2.go\`**:
  - Add \`receiveSubscription\` struct and per-account subscription tracking on \`JsonRpc2Client\`, with refcounting so multiple concurrent websocket subscribers for the same account share one signal-cli subscription.
  - Add \`subscribeReceive\` / \`unsubscribeReceive\` JSON-RPC method wrappers and \`acquireReceiveSubscription\` / \`releaseReceiveSubscription\` refcount helpers.
  - In \`ReceiveData\`, unwrap manual-mode notifications (\`{subscription, result}\`) so downstream consumers see the same envelope shape as in auto mode and don't have to know which mode is in use.
  - Modify \`GetReceiveChannel\` to accept an \`account\` parameter; if non-empty, acquire a subscription before registering the channel. Make the channel buffered (capacity 64) to absorb short subscriber stalls.
  - Modify \`RemoveReceiveChannel\` to look up the channel's account internally and release the subscription on detach.
- **\`src/client/client.go\`**: thread \`number\` through \`SignalClient.GetReceiveChannel(number string)\` so the underlying \`JsonRpc2Client\` knows which account to subscribe.
- **\`src/api/api.go\`**: pass \`number\` to \`GetReceiveChannel\` from \`handleSignalReceive\`.

## Backward compatibility

Fully backward compatible:
- When \`JSON_RPC_RECEIVE_MODE\` is unset or set to \`on-start\`, signal-cli runs in auto mode and the new \`subscribeReceive\` code path in \`GetReceiveChannel\` is gated by \`if account != \"\"\` — but in any case, calling \`subscribeReceive\` against an auto-mode daemon would just fail with no harm done, and the existing unsubscribed-broadcast path keeps working.
- The notification unwrap step in \`ReceiveData\` is a no-op for auto-mode notifications because their params do not contain the \`{subscription, result}\` wrapper structure.
- All public API signatures retain compatibility except \`SignalClient.GetReceiveChannel\`, which now takes a \`number string\` argument. The only in-tree caller (\`handleSignalReceive\`) is updated. External callers would need a one-line update.

## Validation

Local validation against a real Signal account, real signal-cli-rest-api container in \`MODE=json-rpc\` mode, and a real bot consumer (nora-ng-2). Test scenario: stop the websocket client, send 10 messages from another Signal account into the group during a 60-second outage, restart the client, count delivered messages.

| Configuration | Result |
|---|---|
| Master (\`on-start\`, default) | **10/10 lost (100% loss)** |
| This PR (\`JSON_RPC_RECEIVE_MODE=manual\`) | **0/10 lost (0% loss)** |

The signal-api log trace clearly shows the new lifecycle:

\`\`\`
20:22:25 Subscribed to receive notifications for account +1XXXXXXXXXX (subscription=0)
20:22:47 Unsubscribed from receive notifications for account +1XXXXXXXXXX (subscription=0)
20:28:32 Subscribed to receive notifications for account +1XXXXXXXXXX (subscription=1)
\`\`\`

Between unsubscribe (20:22:47) and resubscribe (20:28:32), 10 messages were sent into the group. After resubscribe, all 10 were delivered to the websocket client within ~10 seconds.

## Notes for review

- Refcounting assumes \`subscribeReceive\` and \`unsubscribeReceive\` are idempotent at the signal-cli layer. Multiple concurrent subscribers for the same account share one underlying signal-cli subscription. If signal-cli ever changes that contract this would need revisiting.
- The notification format unwrap in \`ReceiveData\` is unconditional but harmless in auto mode (auto-mode \`params\` don't have a \`subscription\`/\`result\` shape, the unmarshal fails silently and \`resp1.Params\` is left untouched).
- The channel buffer bump from unbuffered to capacity-64 in \`GetReceiveChannel\` is unrelated to manual mode; it just provides a small headroom so a subscriber that takes ~milliseconds to drain doesn't cause a single-message drop. Happy to split that into a separate commit if you'd prefer.
- I have NOT updated the README or tests as part of this PR — happy to add those if you'd like the PR to land.

## Open questions

1. Default value: should this PR also flip the default to \`manual\`? I left it opt-in to be conservative, but auto mode is the buggy mode and I'd argue \`manual\` should be the new default (with \`on-start\` opt-in for users who want the old behaviour).
2. Naming: \`JSON_RPC_RECEIVE_MODE\` follows the existing \`JSON_RPC_*\` env-var convention in the helper. Open to a different name if you prefer.
3. Should I cherry-pick this against any active release branches?


src/client/jsonrpc2.go
```diff
@@ -270,16 +296,102 @@ func (r *JsonRpc2Client) ReceiveData(number string, receiveWebhookUrl string) {
...
+func (r *JsonRpc2Client) GetReceiveChannel(account string) (chan JsonRpc2ReceivedMessage, string, error) {
+	c := make(chan JsonRpc2ReceivedMessage, 64)
 
 	channelUuid, err := uuid.NewV4()
 	if err != nil {
 		return c, "", err
 	}
 
+	if account != "" {
+		if err := r.acquireReceiveSubscription(account); err != nil {
```
bbernhard, line 387: I think I don't understand that code part here. Isn't `account` always non-empty? So, I guess we would end up always in this code branch, no matter which mode is selected via `JSON_RPC_RECEIVE_MODE`.

bbernhard (commented):
Thanks for the PR

src/client/jsonrpc2.go
MykolaBalakin: Looking further into the code,  I'd say that even if gin, for any reason, calls this method with an empty string, this method shouldn't be responsible for deciding whether the value of `account` is a valid account. If signal-cli does not return an error for its subscribeReceive call, then `account` is a valid account.

So if @strichter has no objections and there are no other remarks from @bbernhard, I’d push a fix to this comment to get the PR merged and released.

strichter: Thanks!

MykolaBalakin: @strichter, I've created a [PR](https://github.com/strichter/signal-cli-rest-api/pull/1) to your fork just so that the code you've implemented stays as your PR (otherwise I'll need to push a commit into my fork and create a new PR to the bbernard's repo, which will look like I've implemented the actual receive-mode fix).

MykolaBalakin: @strichter, could you please check the [PR](https://github.com/strichter/signal-cli-rest-api/pull/1)? This issue is a blocker for one of my apps, so I want to get your fix merged, but I don't feel that it's fair for me to create my own PR and be shown as the author of the PR with your code.

strichter: TBH, Claude wrote this months ago and I have no special insight into the code here. All I know is that I used my version on prod since April and had no more dropped messages.

Issue #622, opened 2024-11-27
Swagger claims receive endpoint returns string, but it's an object


crummy:
### The problem

I ran the liblab API generator against the swagger code with:

```
liblab init --spec https://bbernhard.github.io/signal-cli-rest-api/src/docs/swagger.json
liblab build --language java
```
(I had to fix the /search endpoint missing {number}, but after that...)

When I call the `/v1/receive` endpoint, I get this error in my client:
```
com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `java.util.ArrayList<java.lang.Object>` from Object value (token `JsonToken.START_OBJECT`)
 at [Source: (String)"[{"envelope":{"source":"+178...","sourceNumber":"+1781...","sourceUuid":"...","sourceName":"...","sourceDevice":2,"timestamp":1732652479395,"dataMessage":{"timestamp":1732652479395,"message":"...",...
```

The function `func (a *Api) Receive(c *gin.Context)` does seem to return a string, rather than a JSON object. But presumably it's a JSON string, returned from the client.

I think the fix would be to, in signal-cli-rest-api, parse the returned string into an object, return it with `c.JSON(200, obj)`, and add the object type to the swagger comments. Does that sound right?

### Are you using the latest released version?

- [X] Yes

### Have you read the troubleshooting page?

- [ ] Yes

### What type of installation are you running?

signal-cli-rest-api Docker Container

### In which mode are you using the docker container?

Native Mode

### What's the architecture of your host system?

x86-64

bbernhard:
Thanks for the report!

> I think the fix would be to, in signal-cli-rest-api, parse the returned string into an object, return it with c.JSON(200, obj), and add the object type to the swagger comments. Does that sound right?

yeah, right. It is on my todo list since a while, but I haven't found time to work on that.

Issue #199, opened 2022-01-17
Swagger description for /v1/receive/{number} return type incorrect?


heinrich-ulbricht:
### The problem

When using NSwag to generate C# code the return type for _/v1/receive/{number}_ is wrong. What gets generated is `ICollection<string>` but should rather be `ICollection<object>` since the returned list does not contain strings but objects. 

Looking at the Swagger description it seems the described return value does not reflect what is actually returned. It seems like `[string]` should be something like `[object]`?

### Are you using the latest released version?

- [X] Yes

### Have you read the troubleshooting page?

- [X] Yes

### What type of installation are you running?

signal-cli-rest-api Docker Container

### In which mode are you using the docker container?

Normal Mode

### What's the architecture of your host system?

x86-64

bbernhard:
Thanks for the ticket!

For the `receive` the results from `signal-cli` are directly returned without parsing them into any datastructures first - that's why internally it's just a big string blob (that's btw. also the reason why the data, that is returned in the `receive` endpoint, is camelCased, whereas everything else uses snake_case). But I want to clean that up with one of the next releases - so `v1/receive` will probably be marked as deprecated and superseded with `v2/receive`. That should also fix the documentation issue :)

heinrich-ulbricht:
The code generator I use (NSwag) parses the result via [Json.NET](https://www.newtonsoft.com/json) into objects. When presented with a [string] it assumes the collection contains strings and expects a " after the opening [. But it finds a { which means it's an object. This throws an exception. So for this kind of code generator it makes a difference if it's [string] or [object].

After finding the cause I changed the type and all was well. But it would be nice if this manual step would go away. And it sounds like with the new `v2/receive` endpoint it might indeed. Thanks for the feedback.

PR #838, opened 2026-04-19
Pin dependencies

ildyria:
Ensure the the CI dependencies are pinned instead of using tagged version. This mitigates attacks such as remove-and-republish on tags.

This is how for example how the compromise of  trivy propagated to multiple npm and python packages:
https://www.aquasec.com/blog/trivy-supply-chain-attack-what-you-need-to-know/

PR #842, opened 2026-05-04
Add swagger definitions for the receive end point

Gara-Dorta:
This PR closes #822, #622 and #199.

As discussed in #822 it generates json-schema files using the signal-cli utility for it and then uses a custom script to embed those in the generated `docs.go` file.

I made the script as simple as possible, which means that it doesn't update the `.yaml` file. So I removed the file from the repo.

I also updated the Readme for the docs.

If https://github.com/swaggo/swag/issues/2165 gets fixed, I'll make a new PR to remove the script and update the annotations accordingly but I think it still makes sense to merge as is and not wait for that.

I used copilot extensively to write this as I'm not so familiar with writing go code. I still reviewed everything and checked that it makes sense but please, have a careful look at the script in case I missed something.


bbernhard:
Thanks for the PR!

I noticed that you do a quite some bit of string parsing to find the correct place in the json/go files. Have you maybe considered unmarshaling the json/go file into a `json.RawMessage` and operating on that? I guess that could maybe reduce the code a bit and make it easier to maintain?

Gara-Dorta:
Thanks for reviewing!

I tried first doing it as you suggest but the marshaling back into json modified the file, as it would create diffs where there weren't any. As far as I recall the modifications were on reordering fields, maybe also on blank spaces.

If you think that's an acceptable trade-off, I'm happy to go down that route.

Gara-Dorta:
As you are not compiling signal-cli anymore in this repo. I temporarily pointed the download of the schemas to my fork of signal-cli. I plan to make new releases for the schemas there until https://github.com/AsamK/signal-cli/pull/2040 gets merged in the main repo.

I considered adding generating the schemas in your new repos but that seemed like a lot more work.

bbernhard:
> Thanks for reviewing!
> 
> I tried first doing it as you suggest but the marshaling back into json modified the file, as it would create diffs where there weren't any. As far as I recall the modifications were on reordering fields, maybe also on blank spaces.
> 
> If you think that's an acceptable trade-off, I'm happy to go down that route.

But I guess that would only be a problem the first time? :thinking: What I am also not sure is whether the automatic generation via the CI is a good idea or whether it would be better to just notify (e.g by failing the build) in case the swagger documentation needs to be manually re-generated. What I am always worried is, that those automations open attack vectors for cleverly crafted (shell) injection attacks or open up threat vectors in case an upstream repository gets compromised.  But maybe I am unnecessarily cautious here.

Gara-Dorta:
Yes, the extra changes would only happen the first time.

I get the worry about the new CI job, there has been so many attacks lately. The only thing I can think of is to pin the versions of the actions, like it was done in #838. You'll be the one creating the releases, so it depends on whether you prefer the extra work of generating the docs and re-run a failing CI or use the job. I like automating as much as possible, so if I were you I'd keep the automation. From my side, I just care that the final docs are generated correctly. I don't think it'll be much work to add a check to the release work to fail if the docs are out of date, if that's what you want.

bbernhard:
> Yes, the extra changes would only happen the first time.
> 
> I get the worry about the new CI job, there has been so many attacks lately. The only thing I can think of is to pin the versions of the actions, like it was done in #838. You'll be the one creating the releases, so it depends on whether you prefer the extra work of generating the docs and re-run a failing CI or use the job. I like automating as much as possible, so if I were you I'd keep the automation. From my side, I just care that the final docs are generated correctly. I don't think it'll be much work to add a check to the release work to fail if the docs are out of date, if that's what you want.

I guess that should be fine then. I think reducing the complexity of the script would be worth it - especially, since I would like to switch to OpenApi 3.x at some point and I guess some changes might be needed then to the script. If it's not easily do-able with golang, I think a Python script should be okay too. 

For the CI: I think I would prefer a simple build failure/CI failure for now. I know, that sounds a bit boring, but I'd like to see how far that gets us. I _think_ the API documentation doesn't change that often, so I think manually a manual rebuild should be fine for now. That way it should be possible to really lock down the permissions of the CI job. :)

Thanks a lot for all the effort you are putting into this - it's very much appreciated!

src/docs/swagger.json
Gara-Dorta, line 1: To check that this file contains the same information but now reordered alphabetically, I wrote a tiny parser to check each key/value. I loaded this file and one I generated from before the changes using commit 82f71512129b1d32603a6b5c695f5287e652b864

src/docs/add_v1_receive_schemas.go
```diff
@@ -0,0 +1,327 @@
...
+	content = strings.ReplaceAll(content, "` + \"`\" + `", "`")
+	content = strings.Replace(content, schemesTemplateValue, `"`+schemesPlaceholderToken+`"`, 1)
+	return content
+}
+
+func encodeForGoRawString(content string) string {
+	content = strings.ReplaceAll(content, "`", "` + \"`\" + `")
+	content = strings.Replace(content, `"`+schemesPlaceholderToken+`"`, schemesTemplateValue, 1)
+	return content
+}
```
Gara-Dorta, line 135: I've simplified the script as requested to marshal/unmarshal the whole json. I still had a to add a few checks in these two functions because the docs.go doesn't contain a fully valid json. So these lines convert it into a valid one and then set it back to how it was.

bbernhard:
many thanks!

PR #844, opened 2026-05-06
feat: deploy pages with CI

Gara-Dorta:
This PR builds the docs on each release via a CI workflow. This has a few advantages over the current approach:

1. The build process is faster because the pages artifact only includes the index.html and the swagger.json instead of the whole repository
3. The pages only contain the changes from the latest release instead of whatever is currently in master
4. The dynamic version setting that I did in #830 only works for local deployments of the page. If you go to https://bbernhard.github.io/signal-cli-rest-api/ it still shows version `1.0`. The approach in this PR modifies the version in the swagger.json so that it is updated also in the github pages.

You will have to change the settings for this work. So in Settings -> Pages -> Build and deployment -> Source change it from `Deploy from branch` to `Github actions`.

When testing this github was a bit funny about it. You might need to manually unpublish and republish the page when doing the switch.

Here is the workflow running on my fork https://github.com/Gara-Dorta/signal-cli-rest-api/actions/runs/25430692982
Here is the generated page https://gara-dorta.github.io/signal-cli-rest-api/

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -1,0 +1,37 @@
+name: Deploy API Docs to GitHub Pages
+
+on:
+  release:
+    types:
+      - published
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup Pages
+        uses: actions/configure-pages@v6
+
+      - name: Build static documentation bundle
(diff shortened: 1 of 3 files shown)


Gara-Dorta:
FYI, I also tried to update the version dynamically using the current approach in master. I got it almost working here https://github.com/bbernhard/signal-cli-rest-api/compare/master...Gara-Dorta:signal-cli-rest-api:swagger-docs-version However, the link to the `swagger.json` ends up as `blob:none:<very long random string>` due to the `URL.createObjectURL` call. While using `spec:` instead of `url:` leads to not having any file at all. So that's why I think that deploying via CI is the best option.

Gara-Dorta:
@bbernhard would you have time soon to review this?

bbernhard:
Thanks!

PR #848, opened 2026-05-09
Add empty permissions block at workflow level

arnehuang:
None of the three workflows currently declare a top-level `permissions:` block, so they inherit the repo default for `GITHUB_TOKEN`. 

It seems like none of them need any GitHub API write scope, since they only push to Docker Hub

This caps the blast radius if `GITHUB_TOKEN` ever leaks.


bbernhard:
Thanks!

PR #849, opened 2026-05-15
Check docs are updated

Gara-Dorta:
This PR adds a check that the docs are updated before making a release. If they are not updated, the release fails.

Here is an example of it failing https://github.com/Gara-Dorta/signal-cli-rest-api/actions/runs/25933007969
Here is an example of it passing https://github.com/Gara-Dorta/signal-cli-rest-api/actions/runs/25932853154


bbernhard:
many thanks!

Gara-Dorta:
@bbernhard in case it wasn't clear. Now that #842 is merged. This check will fail if you try to make a release. I'll make a new PR next week so that the workflow also includes the changes in #842 when checking.

PR #850, opened 2026-05-16
Group member label support

Berglund:
I had a use-case for the new label (and emoji) feature in groups, and noted that `signal-cli` had support (https://github.com/AsamK/signal-cli/commit/27a722dc757eba633f9014738efa0f1c3f722f28) but not `signal-cli-rest-api`. 

If unaware of this functionality, it allows you to set a label for yourself in a group that's then visible for others:
image

This PR has three commits;

[f07c587](https://github.com/bbernhard/signal-cli-rest-api/pull/850/commits/f07c5873099c7d904313c2fc6768d7dcb0bff89e) adds support for setting the label via `PUT` `/v1/groups/<number>/<group id>` with `member_label` and `member_label_emoji`.

[0938262](https://github.com/bbernhard/signal-cli-rest-api/pull/850/commits/0938262814a70af91f7daba84e868586c1c6bc4a) exposes the labels set in various groups on `GET` `/v1/groups/<number>`.

[fc9a90c](https://github.com/bbernhard/signal-cli-rest-api/pull/850/commits/fc9a90cb7704e3729a966730ebc6eb018c9d3b37) updates the public request interface based on review feedback, using the nested structure `{"member_label": {"name": "<label>", "emoji": "<emoji>"}}`.

### DIFF
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -48,12 +48,14 @@ type CreateGroupRequest struct {
 }
 
 type UpdateGroupRequest struct {
-	Base64Avatar   *string              `json:"base64_avatar,omitempty"`
-	Description    *string              `json:"description,omitempty"`
-	Name           *string              `json:"name,omitempty"`
-	ExpirationTime *int                 `json:"expiration_time,omitempty"`
-	GroupLinkState *string              `json:"group_link,omitempty" enums:"disabled,enabled,enabled-with-approval"`
-	Permissions    *ds.GroupPermissions `json:"permissions,omitempty"`
+	Base64Avatar     *string              `json:"base64_avatar,omitempty"`
+	Description      *string              `json:"description,omitempty"`
+	Name             *string              `json:"name,omitempty"`
+	MemberLabel      *string              `json:"member_label,omitempty"`
+	MemberLabelEmoji *string              `json:"member_label_emoji,omitempty"`
+	ExpirationTime   *int                 `json:"expiration_time,omitempty"`
+	GroupLinkState   *string              `json:"group_link,omitempty" enums:"disabled,enabled,enabled-with-approval"`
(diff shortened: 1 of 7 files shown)


src/api/api.go
```diff
@@ -48,12 +48,14 @@ type CreateGroupRequest struct {
...
-	Description    *string              `json:"description,omitempty"`
-	Name           *string              `json:"name,omitempty"`
-	ExpirationTime *int                 `json:"expiration_time,omitempty"`
-	GroupLinkState *string              `json:"group_link,omitempty" enums:"disabled,enabled,enabled-with-approval"`
-	Permissions    *ds.GroupPermissions `json:"permissions,omitempty"`
+	Base64Avatar     *string              `json:"base64_avatar,omitempty"`
+	Description      *string              `json:"description,omitempty"`
+	Name             *string              `json:"name,omitempty"`
+	MemberLabel      *string              `json:"member_label,omitempty"`
+	MemberLabelEmoji *string              `json:"member_label_emoji,omitempty"`
```
bbernhard: Thanks a lot for the PR!

In general I think the PR is fine. What I am unsure though is, the structure of the public interface. I think I'd prefer some like: 

```
MemberLabel struct {
   Name      *string         `json:"name,omitempty"
   Emoji      *string         `json:"emoji,omitempty"`
} `json:"member_label,omitempty"`
```

What do you think?

Berglund: Hey! So sorry for the late response. Parental leave is taking the time it requires 😅 

I like your suggestion! Took a stab at it in https://github.com/bbernhard/signal-cli-rest-api/pull/850/changes/fc9a90cb7704e3729a966730ebc6eb018c9d3b37 - see what you think.

PR #853, opened 2026-05-23
feat: add receive schemas to check-docks worflow

Gara-Dorta:
This PR adds the receive schemas to the check-docs.yml workflow. Here is a successful run of it https://github.com/Gara-Dorta/signal-cli-rest-api/actions/runs/26340800873

This is a follow up of #849 and #842, they got merged at the same time but they have this interaction with each other.

### DIFF
--- a/.github/workflows/check-docs.yml
+++ b/.github/workflows/check-docs.yml
@@ -23,6 +23,15 @@ jobs:
         working-directory: src
         run: go run github.com/swaggo/swag/cmd/swag@v1.16.6 init --requiredByDefault
 
+      - name: Download and inject receive schemas
+        working-directory: src/docs
+        run: |
+          SIGNAL_CLI_VERSION="$(grep -m1 '^ARG SIGNAL_CLI_VERSION=' ../../Dockerfile | cut -d= -f2)"
+          curl -fsSL "https://github.com/Gara-Dorta/signal-cli/releases/download/v${SIGNAL_CLI_VERSION}/signal-cli-${SIGNAL_CLI_VERSION}-json-schemas.tar.gz" -o signal-cli-json-schemas.tar.gz
+          mkdir -p signal-cli-schemas
+          tar xf signal-cli-json-schemas.tar.gz -C signal-cli-schemas
+          go run add_v1_receive_schemas.go signal-cli-schemas
+
       - name: Fail if docs are out of date
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/check-docs.yml
```diff
@@ -23,6 +23,15 @@ jobs:
         working-directory: src
         run: go run github.com/swaggo/swag/cmd/swag@v1.16.6 init --requiredByDefault
 
+      - name: Download and inject receive schemas
+        working-directory: src/docs
+        run: |
+          SIGNAL_CLI_VERSION="$(grep -m1 '^ARG SIGNAL_CLI_VERSION=' ../../Dockerfile | cut -d= -f2)"
```
Gara-Dorta, line 29: I'm not too happy with getting the signal version like this. I think a better approach is to have a separate file in the repo and read it from there. Like how signal-cli has the lib version in this file https://github.com/AsamK/signal-cli/blob/master/libsignal-version

However to keep with single responsibility PRs. If that is something you'd like, I'd do it in a separate one.

bbernhard:
Thanks - looks good to me! But I think it might make sense to keep this PR open until your signal-cli CI PR is merged. :)

Gara-Dorta:
I think so too. However, getting code merged in signal-cli is a slow process. So if you need to make a release in the meantime, I've made new releases in my fork https://github.com/Gara-Dorta/signal-cli/releases so that this workflow and the one #842 don't crash and download the correct json files for each signal-cli version.

bbernhard:
Hmm...what if we check out `signal-cli` and build the json schemas with `./gradlew jsonSchemas` directly in this CI job? :thinking:

Gara-Dorta:
If we are building it ourselves, I think it makes more sense to add it as a new artifact output in https://github.com/bbernhard/signal-cli-native-builds/blob/master/.github/workflows/build.yml than do it here.

When working on #842 I gave a quick try. However, after a few minutes of coding it felt silly to duplicate the work in there when I could simply publish a release in my signal-cli fork by uploading a tag.

bbernhard:
Yeah, the best would be, if the schemas were already publish by signal-cli.

But I'll merge the PR right away, thanks! I'll quickly check how much work it would be to build the json schemas directly in the CI job...maybe it's not that much work. Once signal-cli publishes the schemas, the part can always be reworked.

PR #859, opened 2026-06-05
Expose group membership (isMember) on the /v1/groups endpoints

pinkustar:
first time ever contributing to this repo, so tell me if I am doing something wrong.
I studied the repo and got some help from Opus 4.8, I checked and verified all tests passed and the API endpoint worked locally

## What

signal-cli's `listGroups` output includes an `isMember` flag indicating whether
the local account is still a member of a group. The REST wrapper read this into
`SignalCliGroupEntry` but dropped it when mapping to the public `GroupEntry` /
`ExpandedGroupEntry` types — only `isBlocked` was carried over.

As a result, `GET /v1/groups/{number}` and `GET /v1/groups/{number}/{groupid}`
returned every group signal-cli still knows about — including ones the account
has **left or been removed from** — with no way to distinguish them from groups
you're actively in.

## Change

- Add a `member` boolean to `GroupEntry` and `ExpandedGroupEntry`, mirroring the
  existing `blocked` field, populated from signal-cli's `isMember`.
- Extract the per-entry mapping into `signalCliGroupEntryToExpandedGroupEntry(...)`
  and add a unit test covering `isMember → member` (and the independent
  `isBlocked → blocked`) mapping.
- Update the generated swagger (`docs.go`, `swagger.json`) for `GroupEntry.member`.

The field is **additive** — existing consumers are unaffected. It appears on both
endpoints, in both `expand=false` and `expand=true` responses.

```json
GET /v1/groups/+1XXXXXXXXXX
[
  { "name": "Current Group", "blocked": false, "member": true,  "...": "..." },
  { "name": "Left Group",    "blocked": false, "member": false, "...": "..." }
]

### DIFF
--- a/src/client/client.go
+++ b/src/client/client.go
@@ -119,6 +119,7 @@ type GroupEntry struct {
 	InternalId      string              `json:"internal_id"`
 	Members         []string            `json:"members"`
 	Blocked         bool                `json:"blocked"`
+	Member          bool                `json:"member"`
 	PendingInvites  []string            `json:"pending_invites"`
 	PendingRequests []string            `json:"pending_requests"`
 	InviteLink      string              `json:"invite_link"`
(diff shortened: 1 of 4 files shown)


bbernhard:
Thanks!

PR #867, opened 2026-07-10
Validate message body ranges and harden receive JSON handling

k-bx:
## Summary

Prevent malformed Signal body ranges from being sent and ensure `receive` responses remain valid JSON when `signal-cli` emits logs or diagnostics on stderr.

## Changes

- Validate mentions and text-style ranges against the final message’s UTF-16 length.
- Validate quote mentions against the quoted message.
- Reject quote mentions when no quote message is provided.
- Keep `signal-cli` stderr out of command responses.
- Preserve `INFO` and `WARN` log severity when classifying stderr.
- Treat unprefixed stderr diagnostics, such as stack traces, as warnings.
- Parse newline-delimited receive output as a JSON stream instead of constructing arrays through string concatenation.
- Return a clear error when receive output contains malformed JSON.
- Add regression coverage for Unicode offsets, invalid ranges, streamed messages, and stderr classification.

## Motivation

Signal body-range offsets use UTF-16 code units. Offsets calculated using UTF-8 bytes—especially around emoji and other non-BMP characters—can produce messages that fail validation when received.

Additionally, combining `signal-cli` stdout and stderr could mix routine logs or exception output into JSON responses, making otherwise valid receive output unparsable.

## Testing

```sh
go test ./...
```

### DIFF
--- a/src/client/cli.go
+++ b/src/client/cli.go
@@ -55,10 +55,25 @@ func classifySignalCliOutput(stdout string, stderr string) (string, string, stri
 	output, infoMessages, warnMessages := stripInfoAndWarnMessages(stdout)
 	stderr = strings.TrimSpace(stderr)
 	if stderr != "" {
-		if warnMessages != "" {
-			warnMessages += "\n"
+		stderrOutput, stderrInfoMessages, stderrWarnMessages := stripInfoAndWarnMessages(stderr)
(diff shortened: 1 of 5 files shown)


bbernhard:
Thanks for the PR!

Can you please do me a favor and split that up in smaller PRs - that makes it easier to review.

PR #869, opened 2026-07-14
feat: download the json schemas from the signal-cli releases

Gara-Dorta:
This PR downloads the json schemas from signal-cli releases instead of compiling them. This is possible now that https://github.com/AsamK/signal-cli/pull/2040 is merged and that a new signal-cli release is out that publishes it (v0.14.6).

I tested this by building the image for the current signal cli version and but using 0.14.6 for the schema generation step.

Apologies for the extra work that you had to do to remove the downloading the schemas from my fork from the original PR.


bbernhard:
Thanks a lot!

PR #884, opened 2026-09-07
Add voice_note parameter to the send endpoint

DorskFR:
## What

Adds an optional `voice_note` boolean to `POST /v2/send`, exposing signal-cli's existing `--voice-note` flag.

## Why

signal-cli can mark audio attachments as voice notes, which makes Signal clients render them inline with a waveform and playback scrubber. The REST API had no way to set it, so audio sent through `base64_attachments` always arrives as a generic file attachment.

## How

The plumbing mirrors the existing `view_once` parameter:

- `voice_note` on `SendMessageV2` → `SignalCliSendRequest.VoiceNote`
- json-rpc mode: adds `"voice-note": true` to the `send` request
- native mode: appends `--voice-note` to the signal-cli invocation
- returns `400` if `voice_note` is set without any attachment, matching the `view_once` behaviour

## Testing

- `go build ./...` and `go test ./client ./utils` pass
- Verified in json-rpc mode by capturing the JSON-RPC frames sent to signal-cli: two otherwise identical requests differ only by the presence of `"voice-note": true`, and the flag is absent when the parameter is omitted
- Confirmed end to end against a real account: an AAC attachment sent with `voice_note: true` is delivered as an inline voice note

## Note on the generated docs

`src/docs/docs.go` and `src/docs/swagger.json` were updated by hand (a single `voice_note` property on `api.SendMessageV2`) rather than by re-running `swag init`. A plain regeneration drops the 42 injected `receive.*` definitions added by `add_v1_receive_schemas.go` and reorders the whole document, which would have buried the change. Happy to regenerate properly if you'd prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/api/api.go
+++ b/src/api/api.go
@@ -142,6 +142,7 @@ type SendMessageV2 struct {
 	NotifySelf        *bool               `json:"notify_self,omitempty"`
 	LinkPreview       *ds.LinkPreviewType `json:"link_preview,omitempty"`
 	ViewOnce          *bool               `json:"view_once,omitempty"`
+	VoiceNote         *bool               `json:"voice_note,omitempty"`
 }
 
 type TypingIndicatorRequest struct {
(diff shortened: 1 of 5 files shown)


bbernhard:
Thanks for the PR!

I am not sure though if this should be a user facing API parameter. I think it would be nicer to check the mimetype of the attachment and set the flag accordingly. That way it would happen automatically.

DorskFR:
Thank you for checking the PR
The intent was to make the usage explicit: `--voice-note` is per-message to shape the bubble, not per attachment.
Re-thinking this, audio attachments when they are the only attachment are already detected as sound and the message is shaped as such. 
Let's close this!

